PR #556, opened 2014-04-02
adding support for XSD

runefs:
Only very basic testing has been performed so far E2E. 
Somewhat complex schemas can be parsed with no errors but for complex schemas it's
untested whether the resulting provider parses the input XML correctly.
A simple E2E test has been performed including parsing of XML

### DIFF
--- a/src/Xsd/SimpleSchemaTest.xsd
+++ b/src/Xsd/SimpleSchemaTest.xsd
@@ -1,0 +1,60 @@
+<schema xmlns="http://www.w3.org/2001/XMLSchema" targetNamespace="https://github.com/FSharp.Data/" xmlns:tns="https://github.com/FSharp.Data/" attributeFormDefault="unqualified" >
+  <complexType name="root">
+    <sequence>
+      <element name="elem" type="string" >
+        <annotation>
+          <documentation>This is an identification of the preferred language</documentation>
+        </annotation>
+      </element>
+      <element name="elem1" type="tns:foo" />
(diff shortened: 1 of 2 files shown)


6D65:
What happened to this branch?
Why hasn't it been merged into master? It would be great to have it available from NuGet.

runefs:
For me it stopped when I tried what I could figure out to why the build
failed in something unrelated externally and not locally and there was no
one with a possible cause. I wanted to finish the implementation but have
no actual use for it and with lack of help to find the missing bits. I left
it on pause. I'll gladly resume when time permits if someone could help out
with understanding the build/test problem

2016-07-14 16:34 GMT+02:00 The Dude <notifications@github.com>:

tpetricek:
We should be able to figure out what the build errors are! The first step is probably to update the branch to the latest version....

In general, I'd be quite keen to get this in, even if it does not support all features of XSD. It would be good to have some summary of what is and what is not supported though.

ovatsus:
We think the essential here would be to get someone which actually have a need for this and a good example use that to drive the development

runefs:
Yeah I had the need when I started it several years back but things changed
and I seldom do XSDs any more. I'll help out but if some one could drive it
that would be cool. We also talked about integrating it into the XML
provider (which I personally were/Are against because it didn't fit my
original use case but that should probably be on the to do list

2016-07-18 18:37 GMT+02:00 Gustavo Guerra <notifications@github.com>:

PR #672, opened 2014-08-30
Support for PCL 259

ovatsus:
Work in progress


ovatsus:
Any pointers? @vladima @dsyme

SabotageAndi:
Any idea/timeframe when this is completed?

paulyoung:
I'd love to know this too.

KevinRansom:
There are no plans to add additional profiles at this time. 

There are no plans to update the Type Providers to support the portable class libraries at this time.  I think before we do that we need to do some work on the compiler / typeprovider interface to make type resolution more deterministic and predictable, effectively the type provider is clueless about reference assemblies, type forwarders the equivalence of System.Runtime and mscorlib blah ... blah ... blah ...

As you can probably tell from the threads linked above Don and I differ on what is probably necessary, Don is more of a minimalist than I am, we haven't yet coalesced on a design for what this looks like.  I'm thinking that we will probably tackle that as part of the cross platform coreclr work.  Writing a cross platform type provider is going to require a much more rigorous approach to type binding and so we will need it there for sure.  Once we have that in place if we end up where I want to be, a type provider should not really need to know much about profiles, at least as far as imported types go.

However ... that is a bit down the road.  I'm sorry I can't be more concrete than that, the problem is difficult, if it weren't it would be fixed already.  Fortunately the cross plat work is going to force us to provide a strong solution to it, so eventually we will be in a good place.

Does this help?

Kevin

dsyme:
Hi Kevin,

 I'm all for fixing this by providing further helpers in the fundamental compiler interface as necessary (and then presumably mapping those up through the ProvidedTypes API which all type providers are authored against). Do you have any rough proposal for what this might look like?

 I find it hard to imagine why any reverse-binding logic would be better placed in the compiler rather than in the ProvidedTypes implementation. But that might be because I'm more comfortable making changes to the ProvidedTypes API and we already implement a number of fixups there. In any case you're right that the CoreCLR work may flush this out, at least for CoreCLR targets.

I still do plan to take a look at solving this for F# 4.0 and .NET 4.x in ProvidedTyoes, I've had this red-flagged for ages and I really should have done this, my apologies. I expect a solution there would be useful to CoreCLR as well, even if the compiler is able to provide additional helpers.

KevinRansom:
Hey Don,
On CoreCLR Assembly.Load, Assembly.LoadFrom and Assembly.ReflectOnlyLoad do not work the same, we would have to expose a LoadContext on the API that providedtypes would then need to use to reify reflect only types to import for code generation.

There are a number of reasons why the compiler should control the types that the type provider can use in generated code:

1.In the future FSC and FSI may use AOT (Ahead of Time) compilation the type provider will not be able to grab the fsharp.core that the compiler is using.
2.TypeProviders should be able to support multi-targeting – to do that they need to use the version of fsharp.core that was provided using –r: not typeof<Seq>.Assembly.Location
3.TypeProviders should be able to support profiles – to do that they need to use the fsharp.core that was provided using –r: not typeof<Seq>.Assembly.Location, similarly they need int from System.Runtime or mscorlib and so on for every other runtime type that exist in profiles.
4.We want an abstraction that will work well across platforms, IDEs and build environments.

In short the compiler command line needs to define the complete set of input files required to successfully complete the compilation, the type provider should not need to helpfully hunt around for assemblies to satisfy it's compilation requirements.  The provider should definitely not need to look at configuration files, the registry or environment variables to know what to do.

It seems profoundly wrong to me that a type provider would need code like this, I can’t imagine why a compiler add-in should know so much about the compilation environment.  [](https://github.com/fsharp/FSharp.Data/blob/master/src/CommonProviderImplementation/AssemblyResolver.fs#L15)
````
let private referenceAssembliesPath =  
    if runningOnMono 
    then "/Library/Frameworks/Mono.framework/Versions/CurrentVersion/lib/mono/" 
    else Environment.GetFolderPath Environment.SpecialFolder.ProgramFilesX86  
    ++ "Reference Assemblies"  
    ++ "Microsoft"  
````
This won’t work on coreclr: [](https://github.com/fsharp/FSharp.Data/blob/master/src/CommonProviderImplementation/AssemblyResolver.fs#L76)
````
let private designTimeAssemblies =  
    AppDomain.CurrentDomain.GetAssemblies() 
    |> Seq.map (fun asm -> asm.GetName().Name, asm) 
    // If there are dups, Map.ofSeq will take the last one. When the portable version 
    // is already loaded, it will be the last one and replace the full version on the 
    // map. We don't want that, so we use distinct to only keep the first version of 
    // each assembly (assumes CurrentDomain.GetAssemblies() returns assemblies in 
    // load order, must check if that's also true for Mono) 
    |> Seq.distinctBy fst  
    |> Map.ofSeq
````
In short we need an abstraction that enables a TypeProvider Author to do his work without worrying about how to access the types and that makes it difficult to do the wrong thing.  I don't yet know what it looks like, since I haven't examined enough TPs to figure what would be useful APIs.  

My guess is that whatever we do in ProvidedTypes.fs will be throw-away code, because coreclr has such a different reflection and assembly loading environment.  We can be sure that existing TP's will need a lot of rewriting for coreclr, so many APIs have changed especially since they spend so much code on resolving types and assemblies wrongly.

I hope this makes sense, and I'm sorry I haven't yet a design for you to review

Kevin

dsyme:
Hi Kevin, 

The LoadContext would be immensely useful, that's a fine direction for the core API. 

> It seems profoundly wrong to me that a type provider would need code like this...

I agree with this - the compiler should manage the set of bindable types. And in any solution,  logic like the above gets taken out if all type providers and into the compiler and/or ProvidedTypes API

> ...throwaway code...

Yes eventually the portion of code used for type binding could be replaced, which would be great. But additional code in ProvidedTypes looks needed to rewrite quotations  and provided type definitions to target bindable types   Also any solution for .NET 4.x or F# 4.x is going to be used for a long time - realistically these will dominate F# usage for a good time.

PR #768, opened 2015-01-01
Fix test failing in 2015

nwolverson:
Just ran unit tests for the 1st time and got a single failure

* Heterogeneous types with Nulls, Missing, and "" should return None on all choices

Looks like it's hard-coding 2014 as the current year.


ovatsus:
Tks

Issue #578, opened 2014-04-25
Write API for CsvProvider


ovatsus:

ovatsus:
Already implemented

Issue #752, opened 2014-11-28
[CSV] Allow parsing of single rows


isaacabraham:
Once you've generated a typed CsvProvider, sometimes it is handy to be able to chuck single (or multiple) rows of data at the provider to parse and process. Currently you have to also provide the header row in the string that is passed to the Parse method as well. It would be useful to simply be able to call the Parse method with just the row data without having to provide the headers again e.g.

    type MyDataset = CsvProvider< @"Dataset.csv">
    let row = MyDataset.Parse "Isaac,Abraham,34"

rather than

    let row = MyDataset.Parse "Forename,Surname,Age\r\nIsaac,Abraham,34"

ovatsus:
This should be easy to do, we already have a CsvRow constructor, we can add a CsvRow.Parse method as well.

@nwolverson, do you want to take it?

nwolverson:
Should be able to take a look this week.

Issue #644, opened 2014-06-28
Try the Extensions Member Draft


ovatsus:
Maybe we could have something like adding columns on the CsvProvider

https://visualfsharp.codeplex.com/SourceControl/network/forks/dsyme/cleanup/contribution/6853

PR #769, opened 2015-01-01
Csv write API.

nwolverson:
Added append and replace-rows (`With`) methods to CsvFile, and add constructor to generated csv row type.
Constructor has optional parameters for each column but string types unfortunately default to null rather than "".

As per #578. Not sure if this is the right API, guided by `Filter` etc methods. Other caveat, I wasn't able to get the proper defaults for optional parameters in the generated constructor so that's rather 'hacked in' to avoid a NRE on saving the CSV.

### DIFF
--- a/src/CommonProviderImplementation/ProvidedTypes.fs
+++ b/src/CommonProviderImplementation/ProvidedTypes.fs
@@ -1290,9 +1290,6 @@ type ProvidedTypeDefinition(container:TypeContainer,className : string, baseType
             | SymbolKind.SDArray, [typ] -> 
                 let (t:Type) = ProvidedTypeDefinition.EraseType typ
                 t.MakeArrayType()
-            | SymbolKind.Generic genericTypeDefinition, _ when not genericTypeDefinition.IsGenericTypeDefinition -> 
-                // Unit of measure parameters can match here, but not really generic types.
-                genericTypeDefinition.UnderlyingSystemType
             | SymbolKind.Generic genericTypeDefinition, typeArgs ->
                 let genericArguments =
                   typeArgs
--- a/src/CommonProviderImplementation/ProvidedTypes.fs
+++ b/src/CommonProviderImplementation/ProvidedTypes.fs
@@ -1290,6 +1290,9 @@ type ProvidedTypeDefinition(container:TypeContainer,className : string, baseType
             | SymbolKind.SDArray, [typ] -> 
                 let (t:Type) = ProvidedTypeDefinition.EraseType typ
                 t.MakeArrayType()
+            | SymbolKind.Generic genericTypeDefinition, _ when not genericTypeDefinition.IsGenericTypeDefinition -> 
+                // Unit of measure parameters can match here, but not really generic types.
+                genericTypeDefinition.UnderlyingSystemType
             | SymbolKind.Generic genericTypeDefinition, typeArgs ->
                 let genericArguments =
                   typeArgs
--- a/src/Csv/CsvGenerator.fs
+++ b/src/Csv/CsvGenerator.fs
@@ -18,7 +18,8 @@ module internal CsvTypeBuilder =
     { TypeForTuple : Type
       Property : ProvidedProperty
       Convert: Expr -> Expr
-      ConvertBack: Expr -> Expr }
+      ConvertBack: Expr -> Expr
+      Param : ProvidedParameter }
 
   let generateTypes asm ns typeName (missingValuesStr, cultureStr) replacer inferredFields =
     
@@ -28,7 +29,8 @@ module internal CsvTypeBuilder =
       { TypeForTuple = typWithoutMeasure
         Property = ProvidedProperty(propertyName, typ, GetterCode = fun (Singleton row) -> Expr.TupleGet(row, index))
         Convert = fun rowVarExpr -> conv <@ TextConversions.AsString((%%rowVarExpr:string[]).[index]) @>
-        ConvertBack = fun rowVarExpr -> convBack (Expr.TupleGet(rowVarExpr, index)) } )
+        ConvertBack = fun rowVarExpr -> convBack (Expr.TupleGet(rowVarExpr, index))
+        Param = ProvidedParameter(propertyName, typ) } )
 
     // The erased row type will be a tuple of all the field types (without the units of measure)
     let rowErasedType = 
@@ -36,6 +38,9 @@ module internal CsvTypeBuilder =
       |> replacer.ToRuntime
     
     let rowType = ProvidedTypeDefinition("Row", Some rowErasedType, HideObjectMethods = true)
+
+    let ctor = ProvidedConstructor([ for field in fields -> field.Param ], InvokeCode = (fun args -> Expr.NewTuple(args) ))
+    rowType.AddMember ctor
     
     // Each property of the generated row type will simply be a tuple get
     for field in fields do
@@ -76,4 +81,4 @@ module internal CsvTypeBuilder =
 
       Expr.NewDelegate(delegateType, [rowVar], body)
 
-    csvType, csvErasedTypeWithRowErasedType, stringArrayToRow, rowToStringArray
+    csvType, csvErasedTypeWithRowErasedType, rowType, stringArrayToRow, rowToStringArray
--- a/src/Csv/CsvGenerator.fs
+++ b/src/Csv/CsvGenerator.fs
@@ -30,7 +30,7 @@ module internal CsvTypeBuilder =
         Property = ProvidedProperty(propertyName, typ, GetterCode = fun (Singleton row) -> Expr.TupleGet(row, index))
         Convert = fun rowVarExpr -> conv <@ TextConversions.AsString((%%rowVarExpr:string[]).[index]) @>
         ConvertBack = fun rowVarExpr -> convBack (Expr.TupleGet(rowVarExpr, index))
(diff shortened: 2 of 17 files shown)


ovatsus:
Tks, I'll review this later

colinbull:
I have just had a quick look at the diff's for this and a couple of things spring to mind. 

* When adding a row if the sample does not have null values should we allow this? Or should the provider invalidate and regenerate the types as `Nullable<'a>` or `Option<'a>`

* As I understand it, the `With` method simply creates a completely new CsvFile instance and throws the other instance away. If this is the case the then a better name might be `New` could still be referenced. `With` seems to imply an in-place update, but that's just my opinion.

Looks good thou.

nwolverson:
On the 1st point, not sure if I follow - is the concern with columns with a type such as int? I can see an argument to make the constructor argument non-optional instead of defaulting to 0, not sure what I prefer. To my mind if you wanted to switch the column to nullable you had the wrong sample/schema in the first place (and can fix that). Maybe I misunderstand.

I don't think `With` implies in place update, I think it implies functional update (ala record update), but is a poor fit here - I did consider `WithRows`. As you mentioned `New`, morally this is a `new` operation and it's weird to need an existing file to replace the contents of, but it's troubling to move it from the instance method as it depends on other data (eg headers). It could be done, but as it is it works for both `CsvProvider` and manual use of `CsvFile` (the subclass). 

Could add a static creation method to the `CsvProvider`, or a new generated subtype of `CsvFile<_>` with a constructor but I'm concerned it's unnecessarily complex.

ovatsus:
`Append` makes sense, but I'm not sure about `With`. `With` usually implies a functional update of a field, but here we're completely replacing all the rows, so this looks like a poor man's constructor, doesn't make sense to me to make it a n instance method. I think either a static `New` method or another constructor for CsvProvider would be better suited, and follow more closely the pattern of JsonProvider and XmlProvider

I'd also like to do #752 and #644, so we need to make sure the new methods we add here don't collide with that

src/Csv/CsvGenerator.fs
```diff
@@ -28,7 +29,8 @@ module internal CsvTypeBuilder =
       { TypeForTuple = typWithoutMeasure
         Property = ProvidedProperty(propertyName, typ, GetterCode = fun (Singleton row) -> Expr.TupleGet(row, index))
         Convert = fun rowVarExpr -> conv <@ TextConversions.AsString((%%rowVarExpr:string[]).[index]) @>
-        ConvertBack = fun rowVarExpr -> convBack (Expr.TupleGet(rowVarExpr, index)) } )
+        ConvertBack = fun rowVarExpr -> let backExpr = convBack (Expr.TupleGet(rowVarExpr, index)) in <@@ if (%%backExpr : string) = null then "" else (%%backExpr : string) @@>
```
ovatsus: The null to "" conversion needs to be done on the row constructor instead, to maintain the invariant of not having null as a string value on CsvProvider anywhere

src/Csv/CsvGenerator.fs
```diff
@@ -28,7 +29,8 @@ module internal CsvTypeBuilder =
...
+        Param = ProvidedParameter(propertyName, typWithoutMeasure, optionalValue = None ) } )
```
ovatsus: You can omit `, optionalValue = None` as it's the default

ovatsus: This should be `typ` instead of `typWithoutMeasure`

ovatsus:
I'd also like the row constructor parameters to be all mandatory, not sure why they are getting optional though, as you're not specifying a default value

ovatsus:
Please also regenerate the signature tests as part of this pull request (either using `generateAllExpected` in SignatureTests.ts or using Test.fsx). You'll probably need to rebase before that as the expected files have been changed after this PR was created

nwolverson:
Thanks for the review.

I intended the row constructor parameters to be optional, motivated by a particularly large number of  columns in a CSV, but on reflection I agree mandatory is the right thing. (removing `optionalValue = None` in fact makes them mandatory)

Changed to add a `New` function to the generated Csv type (still not sure should it be a constructor just?), unfortunately generating it seemed to get a bit complicated.

Unfortunately I'm having some problems generating the signature tests, as the units of measure in the constructor are blowing up the pretty-print (`Debug.fs` `getConstructorBody` - `EraseType` sees it as a generic type and fails on this - don't really understand this yet.)

nwolverson:
Now can generate signature tests and everything passes (other than freebase complaining about API keys). Not sure if the ProvidedTypes change here is acceptable.

src/Csv/CsvGenerator.fs
```diff
@@ -28,7 +29,8 @@ module internal CsvTypeBuilder =
       { TypeForTuple = typWithoutMeasure
         Property = ProvidedProperty(propertyName, typ, GetterCode = fun (Singleton row) -> Expr.TupleGet(row, index))
         Convert = fun rowVarExpr -> conv <@ TextConversions.AsString((%%rowVarExpr:string[]).[index]) @>
-        ConvertBack = fun rowVarExpr -> convBack (Expr.TupleGet(rowVarExpr, index)) } )
+        ConvertBack = fun rowVarExpr -> convBack (Expr.TupleGet(rowVarExpr, index))
+        Param = ProvidedParameter(propertyName, typ) } )
```
ovatsus: You should adjust casing of constructor parameter here to camelCase. I think we do that in JsonProvider also

src/Csv/CsvProvider.fs
```diff
@@ -75,15 +75,26 @@ type public CsvProvider(cfg:TypeProviderConfig) as this =
...
+
+
+      let stringArrayToRowVar = Var("stringArrayToRow", stringArrayToRow.Type)
+      let rowToStringArrayVar = Var("rowToStringArray", rowToStringArray.Type)
+      
+      let paramType = (replacer.ToRuntime typedefof<seq<_>>).MakeGenericType(rowType)
+      let headers = match sampleCsv.Headers with 
+            | None -> <@@ None: string[] option @@> 
+            | Some headers -> Expr.NewArray(typeof<string>, headers |> Array.map (fun h -> Expr.Value(h)) |> List.ofArray) |> (fun x-> <@@ Some (%%x : string[]) @@>)
+      let newfn = ProvidedMethod("New", [ ProvidedParameter("rows", paramType) ], csvType, IsStaticMethod = true, InvokeCode = (fun (Singleton paramValue) ->
```
ovatsus: Thinking a bit more about it, I think it's preferable to use a constructor instead of a static method New, for better parity with the other providers

src/CommonProviderImplementation/ProvidedTypes.fs
```diff
@@ -1290,6 +1290,9 @@ type ProvidedTypeDefinition(container:TypeContainer,className : string, baseType
             | SymbolKind.SDArray, [typ] -> 
                 let (t:Type) = ProvidedTypeDefinition.EraseType typ
                 t.MakeArrayType()
+            | SymbolKind.Generic genericTypeDefinition, _ when not genericTypeDefinition.IsGenericTypeDefinition -> 
+                // Unit of measure parameters can match here, but not really generic types.
+                genericTypeDefinition.UnderlyingSystemType
```
ovatsus: Humm, I would prefer not to touch this

src/Csv/CsvRuntime.fs
```diff
@@ -212,6 +212,14 @@ type CsvFile<'RowType> private (rowToStringArray:Func<'RowType,string[]>, dispos
   interface IDisposable with
     member __.Dispose() = disposer.Dispose()
 
+
+
+  /// [omit]
+  [<EditorBrowsableAttribute(EditorBrowsableState.Never)>]
+  [<CompilerMessageAttribute("This method is intended for use in generated code only.", 10001, IsHidden=true, IsError=false)>]
+  static member CreateEmpty (rowToStringArray, rows:seq<'RowType>, headers, numberOfColumns, separators, quote) =    
+    new CsvFile<'RowType>(rowToStringArray, { new IDisposable with member x.Dispose() =() }, rows, headers, numberOfColumns, separators, quote)
```
ovatsus: I think we don't need rowToStringArray here, as we're not parsing but specifying the rows directly

nwolverson: Other direction, this if for writing to a string (Saving). stringArrayToRow is indeed dropped.

nwolverson:
Definitely didn't want to change the UoM related case in the erase in ProvidedTypes, I'll happily say I don't know what I'm doing.

Unfortunately I think this is complaining about the ProvidedParameter itself (its type). The code is this:

            let getConstructorBody (c: ProvidedConstructor) = 
                if c.IsImplicitCtor then Expr.Value(()) else
                seq { for param in c.GetParameters() do yield (ProvidedTypeDefinition.EraseType param.ParameterType) }
                |> Seq.map (fun typ -> Expr.Value(null, typ))
                |> Array.ofSeq
                |> c.GetInvokeCodeInternal false

Thinking maybe pulling it up to `Debug.fs` is a bit more acceptable.

ovatsus:
Merged, sorry for the delay.
I looked more carefully and put back the changes you had to EraseType in ProvidedTypes, they were ok there
Also added a .Map method now that we have a CsvRow constructor, to match the .Filter, .Take, .Skip, and all others

nwolverson:
That's great, thanks.

Issue #57, opened 2013-02-01
Support XSD in XmlProvider


ovatsus:

nwolverson:
Having a look at this as I'd quite like to use it myself.

Struggled to get started at all with using XsdProvider, firstly not finding my XSD (file logic is not as robust as XmlProvider, can at least share this if not merging them), then not parsing simple examples (#771).

Will try to get towards a point I can parse my desired XSD, though I find the format/APIs a bit complex sometimes...

runefs:
Can you share the XSDs that you can't parse?

Mvh
Rune

nwolverson:
Sample here: http://msdn.microsoft.com/en-us/library/dd489283.aspx (#772 and namespace in XSD - seem to have a fix for that https://github.com/nwolverson/FSharp.Data/tree/XsdProvider-targetNamespace - parses with that change + reordering)

http://www.topografix.com/GPX/1/1/gpx.xsd (#772 at design time then namespace in XML at runtime - parses once I reorder and delete the namespace reference).

nwolverson:
Actually I think much of the problems (2 separate issues) are down to incorrect handling of qualified/unqualified element forms, being applied to the schema instead of the instance document (http://www.w3.org/TR/xmlschema-0/#NS) - all quite hard to understand and overly complicated to handle (document default, specified on element, global vs nested then interaction with default ns...)

Think I have an initial fix here so will whip up a PR with tests soonish.

giacomociti:
Hi, in my [hobby project] (https://github.com/giacomociti/AntaniXml) I dealt with`System.Xml.Schema`. So maybe I can help here (but I'm a GitHub and TP noob). My idea is to define a simplified schema model, much like [this] (https://github.com/giacomociti/AntaniXml/blob/master/src/AntaniXml/XsdDomain.fs) but even simpler. Then we can split the task:
* mapping element definitions in xsd files to instances of this simplified xsd model
* mapping instances of this xsd model to `InferedType`.

nwolverson:
That doesn't seem like a bad idea, but have you looked at the current version? By my recollection it was mostly there, maybe some issues with external references & could do with checking by someone who really understands XSD.

And I guess before this can ever be merged, some tidying up of the API as discussed above, some decisions to be made.

giacomociti:
[...]

PR #774, opened 2015-01-06
Sorting out qualified/unqualified forms and namespaces

nwolverson:
Existing code applied default qualified/unqualified element forms to the XSD itself, think they should actually be applied to the target/instance document, see http://www.w3.org/TR/xmlschema-0/#NS

Also tried to reduce some duplication (hopefully not removed too much?), and to respect namespaces in both schema and instance doc. 

I can now parse GPX xsd as mentioned in #57, added a bunch of tests of the namespace stuff.

### DIFF
--- a/src/TypeProviderInstantiation.fs
+++ b/src/TypeProviderInstantiation.fs
@@ -29,6 +29,12 @@ type XmlProviderArgs =
       Culture : string
       ResolutionFolder : string }
 
+type XsdProviderArgs = 
+    { SchemaFile : string
+      ResolutionFolder : string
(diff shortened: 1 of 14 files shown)


ovatsus:
Can you please also add entries to SignatureTests.config so we can test the surface area of the generated API?

nwolverson:
Sure thing. Also noticed that I accidentally included test project changes.

ovatsus:
Thank you.
I see some expressions are not being outputted ok, so there's a bug somewhere on Debug.fd, I'll try to fix that later

PR #781, opened 2015-02-14
Unify types for named XML nodes accross multiple samples

tpetricek:
This is addressing issue [discussed her eon StackOverflow](http://stackoverflow.com/questions/28515305/f-xml-data-parsing-a-subtype/28518022#28518022).

Basically, the global inference does not work on multiple samples, because we were processing the samples separately and then unifying the types. This change runs the global inference on the whole document (with multiple samples), so that the types are the same across multiple samples.

### DIFF
--- a/src/Xml/XmlInference.fs
+++ b/src/Xml/XmlInference.fs
@@ -50,12 +50,17 @@ let getInferedTypeFromValue inferTypesFromValues cultureInfo (element:XElement) 
 /// Infers type for the element, unifying nodes of the same name
 /// accross the entire document (we first get information based
 /// on just attributes and then use a fixed point)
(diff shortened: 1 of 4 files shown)


ovatsus:
LGTM

tpetricek:
The tests failed though - I was looking into that and it looks like something failed in the HTML type provider access to Wikipedia in the docs. Sounds like Wikipedia may be changing their table format? /cc @colinbull

ovatsus:
Yes, I saw that, the failure is unrelated, the docs are going to the live wikipedia page and it changed, was looking at that now

ovatsus:
It's fixed, was just a rename of a column

tpetricek:
Thanks!

Issue #786, opened 2015-02-17
Performance of Csv Type Provider and CsvFile


colinfang:
I have a large standard csv (8 column * 1M rows, around 40MB size). All fields are integers except for 1 column is int -> bool. No missing values or broken separators.

It takes Csv Type Provider 12s to read 100k rows.
It takes CsvFile 2s to read 100k rows.

The 10s time difference is due to type casting & CsvType.Row (like a tuple) construction.

However, in R, using `fread` in `data.table` package I can read the whole 1M rows in 2 sec.

Since the use case only requires to load the file into memory once, is there any performance boost technique I can apply?

I tried `let content = File.ReadAllText(path")` and then use `Parse` method of Csv Type Provider / CsvFile, it doesn't help (of course).

ovatsus:
Well, you can try to run a profiler on it and optimize it, but it's definitely not as optimized as R. Any improvements welcome

tpetricek:
How are you measuring the times? The fact that the type provider needs 25s and CsvFile only 4s sounds a bit odd...

colinfang:
Simply use `...Rows |> Seq.length` and wrap a `Stopwatch` around.

I tried manually to create a type A which mimics the interface of generated `CsvType.Rows` (because its constructor seems private), and type casted the Rows from CsvFile to A

`CsvFile.Load(path).Rows |> Seq.map converterToA |> Seq.length` takes the same time as the normal Csv Type Provider method.

colinfang:
I tested the [CsvReader](http://www.codeproject.com/Articles/9258/A-Fast-CSV-Reader)

If I don't cast, it takes 4 sec to read the whole 1M.

If I cast using column names and store as `List<A>` where I can then use `A.ColumnX`to access field, it takes 40 sec to read the whole 1M rows. (no matter if A is a Class or a Structure)

If I cast using column indices, it takes 6 sec.

So I suspect the internal use of a dictionary in CsvReader that translate column name to index causes the slowness. This might apply to Csv Type Provider as well.

tpetricek:
@colinfang This is a nice piece of research, thanks for looking into this! I suppose it would make sense to do the lookup based on index, rather than based on the column name.

If we wanted to preserve the current behavior (where column names can probably be reordered (?)) then we could probably generate some lookup array where `lookup.[I]` returns the index of the column that had index `i` in the sample. 

That said, simple lookup based on index is probably better option (unless there were any issues related to this in the past?)

[...]

PR #787, opened 2015-02-17
Cache DefaultRemovableAdornerCharacters for performance

colinfang:
See #786 Performance of Csv Type Provider and CsvFile


colinfang:
I expect to see up to 5x speed gain for large csv files.

There is still room to optimize the following.

```{f#}
static member private RemoveAdorners (value:string) = 
    new String(value.ToCharArray() |> Array.filter (not << TextConversions.DefaultRemovableAdornerCharacters.Contains))
```

ovatsus:
Thanks! The property syntax in F# is a bit misleading sometimes, that looked like it was cached already

Issue #754, opened 2014-12-01
HasHeaders option after creating CsvProvider


mofus:
I am using the Csv Provider to parse several large csv files of the same format.

I use a sample file containing the correct headers to create my type provider like so:
`type Measurement = CsvProvider<"sorted100M_sample.csv",CacheRows=false>`

Then I use `Measurement` to load files:
`Measurement.Load("measurements/sorted100MSplit_1")`
These files do not contain headers.

My "problem" is that the first line is skipped on all the subsequent loads.
It would be nice to load with an option to say that the file has no headers.
It is possible when creating the type, but not when loading afterwards.

ovatsus:
@weismat, it should work, please check again in latest master, as I have merged the latest ProvidedTypes.fs

weismat:
I tried to test it using MASTER - received error 
>The type provider constructor has thrown an exception: Type 'FSharp.Data.Runtime.Freebase.FreebaseDataContext' not found"</quote>

ovatsus:
We removed freebase, so that error is weird, I'll double check to see if there's any left overs
In any case if you push your code to your repro, I can have a look

weismat:
I rebuild and the error is gone. But the result is different than I expected.
```
open FSharp.Data
type Test = CsvProvider<"sampleHeaders.csv",CacheRows=false>
[<EntryPoint>]
let main argv = 
    let test = Test.Load("sampleNoHeaders.csv");
    test.Rows |> Seq.iter (fun row -> printf "row:%A" row)
    printfn "%A" argv
```
sampleNoHeaders,csv
1,2
sampleHeaders.csv
Col1,Col2
1,2

PR #788, opened 2015-02-20
Add Parse method to CsvProvider provided Row types

nwolverson:
For #752.

Guess I need to regenerate signature tests but putting this up for comment. In particular, I don't like the implementation, but this seemed more convenient (to put on `CsvFile<>`); and I'm not sure what the best option is for multi-row input. To throw or just take the 1st line?

### DIFF
--- a/src/Csv/CsvProvider.fs
+++ b/src/Csv/CsvProvider.fs
@@ -92,6 +92,12 @@ type public CsvProvider(cfg:TypeProviderConfig) as this =
         let body = csvErasedType?CreateEmpty () (Expr.Var rowToStringArrayVar, paramValue, replacer.ToRuntime headers,  sampleCsv.NumberOfColumns, separators, quote)
         Expr.Let(rowToStringArrayVar, rowToStringArray, body)))
       csvType.AddMember(ctor) 
+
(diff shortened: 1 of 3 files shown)


ovatsus:
Humm, maybe we should have CvsProvider.ParseRows, or CsvProvider.Row.ParseMultiple (same pattern as json), so you don't need that Seq.exactlyOne. But now this seems to be solving the same problem as #754, maybe just doing that would be simpler? We definitely only need one of the two. Not sure which
@isaacabraham @tpetricek what do you think?

nwolverson:
I tend to agree, this seems to be solving the same problem. From #752 the use case was "just the row data without having to provide the headers again". Is the only question whether one might want to parse both with and without headers with the same provider. e.g. to parse a CSV file and append a single CSV line, one could not use with-headers and without-headers provider instantiation and `Append` as the row types would not match. But is that a solution looking for a problem?

ovatsus:
I took your code but moved it to a ParseRows static method in the Csv type itself to be more discoverable and to avoid the exactlyOne exception.
I also changed the implementation so it would still go through the same code that checks separators and ignores empty rows, instead of going directly to the parser

See #807

PR #792, opened 2015-03-02
Use ISO-8601 dates in CSV and XML

danielrbradley:
- Use the built in "Round-trip pattern" as documented here:
   https://msdn.microsoft.com/en-us/library/az4se3k1%28v=vs.110%29.aspx
 - Update tests to match new date-time format.


ovatsus:
You're only updating Xml and Json, can you please also update CSV? https://github.com/fsharp/FSharp.Data/blob/master/src/CommonRuntime/TextRuntime.fs#L91

PR #799, opened 2015-03-12
Fixes occasional StackOverflow in HtmlParser 

colinbull:
This stops an occasional stack-overflow that can happen in the the HTML Parser. Additionally this PR also improves the handling of script tags, especially when there is escaped HTML within those tags.


colinbull:
@ovatsus Do you want to have a quick sanity check of this, before I merge?

ovatsus:
Sorry for the delay. I don't really understand your changes to the parser, but they seem to work :)

Issue #801, opened 2015-03-16
Extend CSV to support fixed column size files?


dsyme:
Hi all,

I was looking at [these samples of intra-day trading prices form the NYSE](ftp://ftp.nyxdata.com/Historical%20Data%20Samples/Daily%20TAQ/) which someone pointed me to.

One of the files containing the master list of stocks contains this.  The first and second columns are fixed size, the third is variable.

I'm wondering if the CSV reader might support this case.  It's not hard to parse manually of course.

This is just a suggestion, feel free to close it pending further requests for this functionality (perhaps it's been requested before too)

```
 00846U1010000A           A              AGILENT TECHNOLOGIES, INC                                   41726005508L 000011999111820051228000000000000000010   00  A           Y114A03307920000001D119991118101  11111                                       
 0138171010000AA          AA             ALCOA INC                                                   41726005508A 000011999010420100420000000000000000010   00  AA          Y130A10696020000001D119510611101  11111                                       
 00768Y2060020AADR        AADR           ADVISORSHARES TR WCM/BNY MELLON FOCUSED GROWTH ARD ETF      400000000    000012010072120091023000000000000000010   05  AADR        N416A00002000000100B720100721000  10010                                       
...
```

colinbull:
I think rather than extending the CSV provider, we should create a new `FixedWidth` provider which has it's own `FixedWidthFile` representation that just reuses the generator and inference from the CSV provider, to provide the types.

tpetricek:
Supporting fixed-width column files would be nice. I guess most of the API from CSV type provider can be reused, but having a separate provider for this sounds like a good idea.

(Unless we want to write an inference that will analyze a file and check if it is fixed-width or CSV - then it would make sense to have just `WhateverMessyFormatProvider` :-))

ovatsus:
CsvProvider is already almost whateverMessyFormalProvider. Ages ago I even suggested renaming it to something TabularTextFormatProvider. I think it would be great if this could just be autodetected, like we already autodetect tab separators instead of comma separators

beekomann:
I don't see how the columns could be auto detected. For example here are the first 71 character columns of the US FedACH Directory.
`011001962O1210003742080312122203950CATHAY BANK                         `
The data columns are positions 1-9, 10, 11-19, 20, 21-26, 27-35, 36-71.

PR #803, opened 2015-03-17
[WIP] Adds support for fixed width file

colinbull:
This adds support for fixed width file formats, the motivation for this comes from #801. It basically adds a `FixedWidthReader` module and then reuses the CSV provider. The original discussion in #801, said to pull this out into another provider, but because the only difference is really the file iterator, so this is now just parameterized see [here](https://github.com/fsharp/FSharp.Data/blob/a81885c72c097c621307fa64289ebc884356e450/src/Csv/CsvRuntime.fs#L310). Otherwise a lot of duplicated code would be produced.

To switch the provider into _FixedWidthMode_ a static parameter has been added.

```
  CsvProvider<"Data/nyse.txt", HasHeaders=false, FixedWidth=true>
```

It would be nice to have more examples of fixed width files that we can test against before we release this, a currently the only one is the small snippet @dsyme gave. 

ToDo:
* [ ] Add more tests and examples besides the NYSE one


beekomann:
Hello Colin

Here are the first few lines of the US FedACH directory:
```
011000015O0110000150020802000000000FEDERAL RESERVE BANK                1000 PEACHTREE ST N.E.              ATLANTA             GA303094470866234568111     
011000028O0110000151072811000000000STATE STREET BANK AND TRUST COMPANY JAB2NW                              N. QUINCY           MA021710000617664240011     
011000138O0110000151101310000000000BANK OF AMERICA, N.A.               8001 VILLA PARK DRIVE               HENRICO             VA232280000800446013511     
011000206O0110000151072505000000000BANK OF AMERICA N.A                 PO BOX 27025                        RICHMOND            VA232617025800446013511     
011000390O0110000151072505000000000BANK OF AMERICA N.A                 PO BOX 27025                        RICHMOND            VA232617025800446013511     
011000536O0110000151111411000000000FEDERAL HOME LOAN BK BOST           9TH FLOOR                           BOSTON              MA021990000617292979011     
011001234O0110000151082613000000000BANK OF NEW YORK MELLON             500 ROSS ST., 154-0960              PITTSBURGH          PA152620000412236333811     
011001276O0110000151093014000000000ONEUNITED BANK                      SUITE 600                           BOSTON              MA021100000877663864811     
011001331O0110000151040315000000000CITIZENS BANK NA                    ROP440                              RIVERSIDE           RI029150000800883422411     
011001726O0110000151091213000000000THE FIRST NATIONAL BANK OF IPSWICH  625 GEORGE WASHINGTON HWY           LINCOLN             RI028650000401574192011     
011001742O0110000151072505000000000BANK OF AMERICA, N.A.               PO BOX 27025                        RICHMOND            VA232617025800446013511     
```

The complete file as well as the file definition can be downloaded from this page https://www.frbservices.org/EPaymentsDirectory/download.html

beekomann:
Based on this 
`CsvProvider<"Data/nyse.txt", HasHeaders=false, FixedWidth=true>`
It appears you plan on inferring the data columns from the sample data. Consider the first 35 character columns of the US FedACH Directory.
`011001962O1210003742080312122203950`
The data columns are positions 1-9, 10, 11-19, 20, 21-26, 27-35. It seems that without a hint regarding the column sizes it will be very tricky to pull those out.

colinbull:
This is probably building a case for not just extending the CsvProvider but adding a new `FixedWidthProvider` that has a static parameter that takes a set of column indices. So in the case above it would look something like, `FixedWidthProvider<"Data/nyse.txt", HasHeaders=false, ColumnIndices="1,10,20,27,35....">` 

This seems to me like a better solution than adding yet another static parameter to the `CsvProvider`.

beekomann:
I think the CSV provider should be split into two parts.
- a new think `TokenizedRowsProvider` which rather than taking a data file takes a collection of rows consisting of the strings making up the columns
- the CSV provider itself which takes a data file and behaves as it does today making use of the `TokenizedRowsProvider`

I have no idea how easy this would be to do. A brute force approach is have a wrapper around CSV provider that simply inserts separators and then uses the CSV provider. 

Here is one motivation for the breaking it up approach. I frequently deal with "mainframe" style files (I can't recall the actual name of these). They are typically structured as follows:
```
File header record
   Batch header record
      Data records
      ...
   Batch footer record
   Batch header record
      Data records
      ...
   Batch footer record
File footer record
```
Each record (header, footer or data) starts with record type marker. The records are fixed length with fixed width columns. There may or may not be line delimiters. There may or may not be blocking filling out the length of the file to a fixed multiple of bytes. 

Extending CSV provider is obviously not the way to go. But it would be very useful to have a toolbox of composable file-reader parts to help build a provider
* something to break the file into records
* something to break the records into fields
* something to infer types from the fields
* etc

beekomann:
Of course a brute force approach is have a wrapper around CSV provider that simply inserts comma's and quotes existing comma's  and then uses the CSV provider.

smoothdeveloper:
I'd have a great usage of this (for same style of file @jbeeko is talking about), I'd like to be able to specify the column names (there are none in the file themselves) and possibly the size of each column myself.

```
52750   10  14330 11136 11156 11176       11961                                         
52750   2   19043 16691 16697 16704       17971 0.1111  0.2111  0.32    0.422   0.5111  
```

the format of this is (it doesn't really matter exactly how one would pass the format to the provider or if type inference can be overriden):

```
col1	i8
col2	i4
col3	f6.0
col4	f6.0
col5	f6.0
col6	f6.0
col7	f6.0
col8	f6.0
col9	f8.0
col10	f8.0
col11	f8.0
col12	f8.0
col13	f8.0
```
The idea of header/body/footer is worth keeping in mind for future extension but I think this logic can remain an external concern for now until the "fixed width" stuff is working great in production :smile:

smoothdeveloper:
Note that I'd need that feature to handle format without separator (fixed width means I'd specify the width of the columns and it can't be inferred with the text file alone).

Is this going to be supported as well? It seems like we just need to be able to give the length of each column in the schema string.

dsyme:
@colinbull Maybe bring this up-to-date?

ovatsus:
I've updated this to the current master. @colinbull , can you enable https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ so I can push?

ovatsus:
@colinbull, sent a PR to your branch

ovatsus:
@colinbull , can you add a bit more of tests and/or examples to the docs besides the NYSE one?

dsyme:
Closing very old PR - please resubmit if you want to revive

PR #808, opened 2015-03-22
Optimize conversion of resizable array to native array

danielrbradley:
- Simple tweak saves 47% of work in certain tests - we basically avoid
building two identical arrays.

There is some other stuff I'm working on that I'll try to get into a decent, tested form to submit, but this is by far the biggest win with no backward compatibility issues.

### DIFF
--- a/src/Json/JsonValue.fs
+++ b/src/Json/JsonValue.fs
@@ -297,7 +297,7 @@ type private JsonParser(jsonText:string, cultureInfo, tolerateErrors) =
             parseEllipsis() // tolerate ... or {...}
         ensure(i < s.Length && s.[i] = '}')
         i <- i + 1
-        JsonValue.Record(pairs |> Array.ofSeq)
+        JsonValue.Record(pairs.ToArray())
 
     and parseArray() =
         ensure(i < s.Length && s.[i] = '[')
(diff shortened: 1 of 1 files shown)


forki:
ouch. This is faster? I wonder if we can tweak such things in FSharp.Core. /cc @dsyme

danielrbradley:
Yup - Seq.toArray doesn't seem to have a check for if the source is a resizable array and return directly rather than iterate and build a new array. Agree it might make sense to also add to core assuming that the type check isn't too expensive.

ovatsus:
Thanks!
I'm still open to changing the array to a list if that makes the parsing measurably faster

danielrbradley:
The list change had some other benefits in certain circumstances, with some extra optimisations too, but it's not 100% consistent improvements, depending on the data, so not happy pushing that forward yet.

dsyme:
@forki or @danielrbradley - yes, please perf test Seq.toArray v. resizeArray.ToArray(), make appropriate fixes in this code https://github.com/Microsoft/visualfsharp/blob/34f300dcaa32a404ede70a3d5f19cc5f3b3022d2/src/fsharp/FSharp.Core/seq.fs#L1231 and submit to https://github.com/Microsoft/visualfsharp.

We should definitely at least add a match case for ResizeArray.

Mind you, calling resizeArray.ToArray() will still always be faster if the input is actually a ResizeArray, since you will avoid the oher checks in Seq.toArray

PR #815, opened 2015-04-10
Convenience function for Basic Auth

tonyj444:
I've created a convenience function for sending http requests with Basic Auth.  Usage will look like:
```
Http.RequestString
  ( "http://httpbin.org/post", 
    headers = [ BasicAuth "user" "password" ],
    body = TextRequest """ {"test": 42} """)
```

I'm not 100% convinced this belongs in the HttpRequestHeaders module, since it actually uses the Authorization header that is already in there.  An alternative would be to have a function that simply returns the encoded string and leaves it to the consumer to pipe this into the Authorization header, something like
```
Http.RequestString
  ( "http://httpbin.org/post", 
    headers = [ BasicAuth "user" "password" |> Authorization ],
    body = TextRequest """ {"test": 42} """)
```
I'm not sure exactly where such a function would belong however.

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -87,6 +87,12 @@ module HttpRequestHeaders =
     let Allow (methods:string) = "Allow", methods
     /// Authentication credentials for HTTP authentication
     let Authorization (credentials:string) = "Authorization", credentials
+    /// Authentication header using Basic Auth encoding
+    let BasicAuth (username:string) (password:string) = 
+        let base64Encode (s:string) = 
+            let bytes = Encoding.UTF8.GetBytes(s)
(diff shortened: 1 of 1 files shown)


tonyj444:
Actually, I've just realised this is probably better achieved by writing directly to the *HttpWebRequest.Credentials* property via *customizeHttpRequest*.

Which I think actually makes this PR redundant

ovatsus:
I actually think this is quite useful, and I think it's on the right place.
Can you just add an example to the docs in Http.fsx before I merge?

tonyj444:
Sure, I'll try and make some time this evening

ovatsus:
ping?

PR #816, opened 2015-04-19
Remove pin on old FsCheck version and update all dependencies

ovatsus:


ovatsus:
@forki, I had to upgrade the test project to .Net 4.5 to be able to use latest FsCheck. I also updated the fake script to use 4.5, but I'm getting a weird error when running the tests via Fake on this PR:

image

Running the tests inside FSI work fine, do you have any idea what could be the problem?

forki:
Oupsis. Are there different fscheck.dlls? Will try to look into this. Maybe I find some time this evening.

ovatsus:
No, just one FsCheck. BTW, running with TestDriven.Net inside VS also gives the same error...

forki:
Mhm. Wondering @kurtschelfthout has seen this before.

ovatsus:
This only seems to fail on Windows, on mono the tests pass

forki:
My gut feeling is: this is another instancr

ovatsus:
Humm, actually appVeyor is failing for some other reason, this test pass, so it seems it's only a problem on my machine. Maybe because I also have VS2015 installed, and the fake command line picks up F# 4.0 instead of F# 3.1?

ovatsus:
It does work on FSI on both VS2013 and VS2015

forki:
Yep that might be an issue

ovatsus:
I'll try to uninstall F# 4.0 and VS2015 to see if it fixes the issue

kurtschelfthout:
Definitely all the symptoms of some FSharp.Core compatibility issue. Try adding app.config redirects?

ovatsus:
Ok, I think the problem is that FSharp.Data.Tests used 4.3.0.0, and FsCheck used 4.3.1.0

ovatsus:
Problem solved, thanks

PR #820, opened 2015-04-22
Use content type encoding

### LINKED ISSUE #819: PUT and POST request don't respect ContentType charset
```f#
let! res = Http.AsyncRequest (url, 
                              headers=[("Content-Type", "application/json; charset=utf-8")],
                              httpMethod = "POST",
                              body = httpRequestBody)
```

Still uses the default encoding set [here](https://github.com/fsharp/FSharp.Data/blob/master/src/Net/Http.fs#L306). 

It should use the charset shouldn't it?

albertjan:
Fixes #819

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -688,7 +688,9 @@ module private HttpHelpers =
 /// Utilities for working with network via HTTP. Includes methods for downloading 
 /// resources with specified headers, query parameters and HTTP body
 [<AbstractClass>]
-type Http private() = 
+type Http private() =
+
+    static let charsetRegex = Regex("charset=([^;\s]*)", RegexOptions.Singleline)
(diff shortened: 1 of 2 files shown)


albertjan:
I don't think the appveyor build fails because of anything I did. :smile:

ovatsus:
Thanks, sorry for the delay, had forgot about this :(

albertjan:
No worries, the work around was easy enough :smile:. Happy this landed though. And in one go :smile:.

ovatsus:
Reverted 2nd commit as it was causing merge issues with other PR's

albertjan:
Ah ok sure an overzealous editors did that :smile: there is an open statement in there that you'll need. 

Sorry this causes issues.

PR #822, opened 2015-05-01
Use ApplyPolicy with ReflectionOnlyLoad based on name (improve #821)

tpetricek:
A better fix for the #821 issue, as suggested by @veikkoeeva. I tested this on my Azure VM and it worked fine.

### DIFF
--- a/src/CommonProviderImplementation/AssemblyResolver.fs
+++ b/src/CommonProviderImplementation/AssemblyResolver.fs
@@ -115,7 +115,9 @@ let private getAssembly (asmName:AssemblyName) reflectionOnly fsharpDataPaths =
     match asm with 
     | Some asm -> asm
     | None -> 
-        if reflectionOnly then Assembly.ReflectionOnlyLoad asmName.FullName
+        if reflectionOnly then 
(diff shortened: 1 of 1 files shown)


tpetricek:
BTW: What are the current release plans for F# Data? It would be nice to get new releases out reasonably frequently. Having this released for the [May 14 hands-on-session in London](http://www.meetup.com/FSharpLondon/events/222161266/) would be nice... :-) 

I'm not yet 100% sure if I'll need this (I was debugging this on Azure VM), but I don't expect the machines running Azure Web apps would have `FSharp.Core.dll` in the GAC....

tpetricek:
(Also, please let me know if I actually broke the build, or if this is caused by some earlier changes..!)

ovatsus:
That failure was just SourceLink, which always fails on PRs, this looks good

ovatsus:
I'll do a release with this soon

tpetricek:
Wonderful, thanks a lot!

ovatsus:
Published

tpetricek:
:+1:

Issue #823, opened 2015-05-01
Object reference not set to an instance of an object


kseo:
The following code no longer compiles after upgrading Mono version to 4.0.0.

    open FSharp.Data
    type Message = JsonProvider<""" {"Timestamp":"2015-04-06T20:17:13"} """>

/Users/kseo/Projects/FSharpDataTest/FSharpDataTest/Program.fs(6,16): error FS3033: The type provider 'ProviderImplementation.JsonProvider' reported an error: Object reference not set to an instance of an object

dmaxime:
I got a very similar error with the CsvProvider on Mono 4.0. Using a tab-delimited file (sample.tsv) containing a date column: 
```
id	request_date
1061	2014-12-09
1145	2014-12-09
1146	2014-12-09
1062	2014-12-09
1147	2014-12-09
1148	2014-12-09
1063	2014-12-09
1149	2014-12-09
1064	2014-12-09
```

The following code breaks
```
open FSharp.Data
let sample = CsvProvider<"sample.tsv", HasHeaders=true >.GetSample()
```
error FS3033: The type provider 'ProviderImplementation.CsvProvider' reported an error: Object reference not set to an instance of an object


while this one works fine:
```
open FSharp.Data
let sample = CsvProvider<"sample.tsv", HasHeaders=true, Schema="request_date=date" >.GetSample()
```

It seems the CsvProvider has issues guessing date column types on Mono. Specifying the `date` type in the schema solves the issue.

ovatsus:
Can you attach debugger and check where the null is coming from?

dmaxime:
Code does not even compile and I do not know how to attach a debugger to
the compilation process. Would you have any link showing how to do this?

Or is there a way to make this code compile and have the error appear at
the execution time?

2015-05-11 16:40 GMT-07:00 Gustavo Guerra <notifications@github.com>:

andrevdm:
I can replicate with this example
    ```F#
    type t = JsonProvider<"""{"d":"2015-05-19T10:00:00.000Z"}""">
    ````
I'll try debug ASAP

brianary:
I'm unable to use FSharp.Data at all in Visual Studio 2013 Update 4.

Running:

    #I @"C:\tools\FSharp.Data\lib\net40"
    #r "FSharp.Data"

Results in:

    Microsoft (R) F# Interactive version 12.0.30815.0
    Copyright (c) Microsoft Corporation. All Rights Reserved.
    
[...]

Issue #426, opened 2014-02-20
[Mono] The type provider 'ProviderImplementation.CsvProvider' reported an error: 1Parameter name: era


ovatsus:
@enricosada, can you take a look please? It's a shame there's no stack trace coming out from type providers

PR #836, opened 2015-05-31
Fix for GetAbbreviatedEraName on mono with .net version +4.5

andrevdm:
This method was already using reflection. This fix checks if
cultureInfo.DateTimeFormat.GetAbbreviatedEraName(era) exists and
if so uses it rather than cultureInfo.Calendar.AbbreviatedEraNames

This fixes https://github.com/fsharp/FSharp.Data/issues/823

### DIFF
--- a/src/CommonRuntime/StructuralInference.fs
+++ b/src/CommonRuntime/StructuralInference.fs
@@ -253,13 +253,38 @@ let inferPrimitiveType (cultureInfo:CultureInfo) (value : string) =
 #else
     let runningOnMono = Type.GetType("Mono.Runtime") <> null
     if runningOnMono then
-      let abbreviatedEraNames = cultureInfo.Calendar.GetType().GetProperty("AbbreviatedEraNames", Reflection.BindingFlags.Instance ||| Reflection.BindingFlags.NonPublic).GetValue(cultureInfo.Calendar, [| |]) :?> string[]
-      let eraIndex =
-        match era with
-        | 0 -> (abbreviatedEraNames |> Array.length) // 0 mean current, last of array
-        | x when x > 0 && x <= abbreviatedEraNames.Length -> era
-        | invalid -> failwith (sprintf "invalid era %i (culture = '%s')" invalid cultureInfo.NativeName)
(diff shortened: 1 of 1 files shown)


andrevdm:
The AppVeyor error does not seem related to this change. The error is 
"Could not load file or assembly 'FSharp.Core, Version=4.4.0.0"

andrevdm:
For the record this fix uses reflection as I don't want to make assumptions about what is available on different .net versions across windows and mono (and I can't reliably test that either). Its not as clear as I would have liked but the existing method already used reflection to get around the version differences. So I don't think this is adding much complexity

ovatsus:
It was already using reflection as the alternative way of doing it was not working on Mono before, they had a big somewhere, can't remember exactly now

andrevdm:
@ovatsus Perhaps the off-by-one bug, referenced here? https://github.com/fsharp/FSharp.Data/issues/426

ovatsus:
Yes. That was it. Maybe if that bug was fixed on latest Mono we can just remove the if altogether and not use reflection?

andrevdm:
That would be better, yes. My concern would be for people using older versions of mono. Perhaps there is some way to handle that from the code. I'll take a look ASAP and see if I can think of something

andrevdm:
Ok... I can't see a better way :(. 

I find it strange that there is a call to "cultureInfo.DateTimeFormat.GetAbbreviatedEraName(era)" in the "else" branch of the "if runningOnMono" as GetAbbreviatedEraNames is only available for .net 4.5 and 4.6 [1]. 

The simplest way would be to remove the mono check and use GetAbbreviatedEraNames but then you need +4.5. So I think the reflection approach I used here is the safest as it will work on both old and new versions of mono. 

Lets hear what everyone else says? But I'm happy with this unless there are any objections.

[1] https://msdn.microsoft.com/en-us/library/system.globalization.datetimeformatinfo.getabbreviatederaname%28v=vs.110%29.aspx

tpetricek:
I submitted a fix for this in #840

PR #840, opened 2015-06-06
Workaround for an era workaround for mono 4

tpetricek:
In mono 4, the workaround is not only no longer needed, but it actually does not work.
This is the most conservative thing I could come up with - try the call normally
(works on .NET and Mono 4+) and if that fails with the Mono 3 error, do the workaround!


tpetricek:
F# Data does not work on Mono 4.0.1 (and possibly also 4.0.0) if you have dates in your sample CSV/XML/JSON files. This fixes the problem.

tpetricek:
Oops, I apparently missed the existing threads about this!

tpetricek:
I would like to do a new FsLab release soon. Any chance to have F# Data with this fix released soon :-)?

ovatsus:
Just released it

tpetricek:
Perfect timing, I now have new release of R provider + Deedle ready to ship too :-) Thank you!

PR #842, opened 2015-06-07
[WIP] Use the built-in mono support in Travis

tpetricek:
It seems that the mono support in Travis is now running using Mono 4.

 - See the log [generated here](https://travis-ci.org/fsprojects/FSharp.Control.AsyncSeq)
 - Which is done [using this Travis config](https://github.com/fsprojects/FSharp.Control.AsyncSeq/blob/master/.travis.yml)

This is an attempt to use the same mechanism for F# Data - sending a PR to see what happens when the Travis build is triggered like this!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,8 @@
-language: objective-c
+language: csharp
 
-env:
-  global:
-    - MONO_VERSION=3.10.0
+sudo: false
 
-install:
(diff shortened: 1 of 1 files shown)


tpetricek:
Well, this fails. /cc @enricosada for expert help :-)

ovatsus:
Fixed as part of #875

Issue #834, opened 2015-05-30
IndexOutOfRangeException for Expires cookie with no comma


Jeff-Cortese:
When calling any of the `Request` functions, an `IndexOutOfRangeException` is thrown when a response has a `"Set-Cookie"` header with an `Expires` cookie that does _not_ have a comma.

For example:
`"s=123xyz; path=/; HttpOnly; Expires=2015-05-30T19:05:43.606-05:00"`

Ideally, in the case where a server uses a different date format, this cookie string would still be parsed.

ovatsus:
A pull request would be very welcome :)

ovatsus:
ping

dsyme:
Closing as analyzed and verified fixed to first approximation

PR #848, opened 2015-06-18
Update Http.fs with more cookie changes

xenocons:
I am not sure what this could break, but basically its just relaxing how the cookie parsing works, this might be a band-aid, but it stops an exception when using SubString with -1

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -671,9 +671,14 @@ module private HttpHelpers =
             |> Array.iteri (fun i cookiePart ->
                 let cookiePart = cookiePart.Trim()
                 if i = 0 then
-                    let firstEqual = cookiePart.IndexOf "="
-                    cookie.Name <- cookiePart.Substring(0, firstEqual)
-                    cookie.Value <- cookiePart.Substring(firstEqual + 1)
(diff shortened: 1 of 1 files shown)


ovatsus:
Does this fix #834?

xenocons:
@ovatsus do you have expected outcomes for that one? i just ran using his output and i didn't seem to get the exception, if there is another test case i can run it

xenocons:
Ok, i've done some more testing (see hair pulling), and i am currently doing this, http+sessions is really just the gift that keeps on giving:
```
            let setCookieThrowaway header = 
                try
                    // websites do silly things, .NET cookieCutter is fairly RFC Strict.
                    cookieContainer.SetCookies( resp.ResponseUri,header )
                with ex -> 
#if DEBUG
                    printfn "invalid cookie set attempt, %s" ex.Message
#else
                    ()
#endif

            // only return headers that are not cookies
            let headers = 
                [ for header in resp.Headers.AllKeys do
                    if header = HttpResponseHeaders.SetCookie then 
                        setCookieThrowaway resp.Headers.[header]
                    else
                        yield header, resp.Headers.[header] 
                        
                        ]
                |> Map.ofList
```

Issue #849, opened 2015-06-25
HTML Type Provider fails with "more than one nested type"


ptrelford:
I'm trying to use the HTML type provider over the the Dr Who page on Wikipedia:

    type who = FSharp.Data.HtmlProvider<"http://en.wikipedia.org/wiki/Doctor_Who">

When I type `who` dot I get a note saying "There is more than one nested type called 'ReferenceWebsites'".

colinbull:
Looks like I have missed a use of the unique name generator somewhere. I'll look into this over the weekend.

-----Original Message-----

colinbull:
Is this on the nuget package or from source?

ptrelford:
Sorry, forgot to mention that, it was from the Nuget package, version 2.2.3

ovatsus:
There's a table and list wit the same name, fixing. Also found that the list values weren't correct when there are links

PR #850, opened 2015-06-28
Fix parsing of html pages with tables and lists with the same name (c…

ovatsus:
…loses #849) + Fix parsing of lists with links + Fix space normalization issues


ovatsus:
@colinbull, can you please review?

colinbull:
Yep, looks good. Does the source link normally, fail for appveyor?

ovatsus:
Yep, it always fails on Pull Requests, because is comparing the checksums with master

PR #857, opened 2015-07-04
Fix parsing of HTML documents with missing closing tags

ovatsus:


ovatsus:
@colinbull can you review?

PR #868, opened 2015-08-19
Reworked the xsdprovider code

runefs:
I've reworked the original Xsd code since my original work was frankly a mess. It's somewhat cleaner now and more importantly uses compiled schema sets which should make for a more correct interpretation of the schemas.


ovatsus:
Sorry for the delay. Merged this into the XsdProvider branch. All the tests are failing with a "key not present in the dictionary", so I might have messed up something. Can you have a look?

PR #873, opened 2015-09-14
Fix HtmlInference inferListType when passing an empty seq

klettier:
When use createDefinitionListType function with description list (dl) containing no value the inferListType function throw an ArgumentException on "Seq.reduce" because input sequence is empty
Here is a simple Html that reproduce the bug

    <html xmlns="http://www.w3.org/1999/xhtml">
      <body>
        <dl>
          <dt><a id="l1" title="My link" href="http://foo.com" target="_blank">Foo</a></dt>
          <dt><a id="l2" title="My link" href="http://foo.com" target="_blank">Foo2</a></dt>
        </dl>
      </body>
    </html>

### DIFF
--- a/src/Html/HtmlInference.fs
+++ b/src/Html/HtmlInference.fs
@@ -38,14 +38,16 @@ let inferHeaders parameters (rows : string [][]) =
 
 let inferListType parameters values = 
 
-    let inferedtype value = 
-        // If there's only whitespace, treat it as a missing value and not as a string
-        if String.IsNullOrWhiteSpace value || value = "&nbsp;" || value = "&nbsp" then InferedType.Null
-        // Explicit missing values (NaN, NA, etc.) will be treated as float unless the preferOptionals is set to true
-        elif Array.exists ((=) <| value.Trim()) parameters.MissingValues then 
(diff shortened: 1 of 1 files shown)


tpetricek:
I'm not sure how this works in the HTML type provider, but the type of the element of an empty list should generally be `InferedType.Top`.

(The idea is that when you have a list of `Top` types and a list of _anything else_ and you're trying to find a common type, the _other type_ always wins. If you had `Null` and _anything else_, then the result is the _other type_ but marked as nullable - this wouldn't be correct, because there is no `null` in the sample.)

So, returning `Top` in the `else` case would logically be more correct. But it might require additional handling somewhere else in the HTML provider code. Could you try changing it and see what happens?

tpetricek:
And I forgot to say, thanks a lot for looking into the issue and sending a PR with a fix :-)!

klettier:
Thank you for your comment, I tested with both values and as far I can see the behavior are equivalent
(I can access the links and their attributes in both cases)
However I'm not an expert and to be honest it is the first time I'm using "HtmlProvider"

klettier:
So, do I need to change my code or the PR can me merged ?

Issue #870, opened 2015-08-20
Single column CSV Row construction gets invalid cast to 1-tuple.


philderbeast:
With FSharp.Data v2.2.0 and v2.2.5 I got this exception when constructing single column CSV rows. I had no such problem with multicolumn CSV rows.

    type S = CsvProvider<"""
    s (string)
    aa
    bb
    """>

    ["aa"; "bb"] |> List.map (fun x -> S.Row(x))
    // System.InvalidCastException: Unable to cast object of type 'System.String' to type 'System.Tuple`1[System.String]'.

I get the same error with ...

    ["aa"; "bb"] |> List.map (fun x -> S.Row x)

philderbeast:
A workaround is to use a dummy 2nd column ...

    type S = CsvProvider<"""
    s (string), dummy (string option)
    aa,
    bb,
    """

    ["aa"; "bb"] |> List.map (fun x -> S.Row(x, None))

ovatsus:
This won't be easy to fix, and it doesn't seem a very common scenario. Not sure if we should just throw a mote explicit exception like "one column csv's not supported

Issue #300, opened 2013-12-24
Review FileSystemWatcher changes


ovatsus:
I'm getting a lot more "background operation" VS dialogs with latest FSharp.Data. Check if this is related to latest changes

//cc @vasily-kirichenko

PR #875, opened 2015-09-21
Work in Progress: Multi-targeting to 4.0, Profile 47, Profile 7 and Profile 259

dsyme:
This is work in progress

I've been looking at what's really required to fix the multi-targeting problem for type providers, using FSharp.Data as the test bed.

As Kevin Ransom has pointed out, the F# compiler really should be responsible for type binding against the referenced set of assemblies.  Short of this, we are using type/quotation rewriting to retarget them to the referenced set of assemblies. 

There are two problems: 

1. Finding the set of referenced assemblies accurately

2. Reflecting over those assemblies

Problem 1 is addressed in this PR via an awful hack to look through the closures in the TypeProviderConfig to find the actual set of referenced assemblies. I believe this functions correctly for all published editions of Visual F# Tools and tooling built using any current edition of the FSharp.Compiler.Services nuget package.

Prior to this PR, problem 2 was addressed in the FSharp.Data CommonProviderImplementation through the use of ReflectionOnly assemblies. However, these are a complete mess on .NET 4.x - with severe limitations such as requiring the use of a global assembly resolve handler, and you can't ReflectionOnlyLoad assemblies that have already been loaded using normal loading. Supposedly .NET 5.x will improve things with a full implementation of a reflection-only binding context for assemblies which is fully independent.

However, I've come to the opinion that, at least in the interim, those interested in implementing cross-targeting type providers really just have to "own our own destiny" and implement proper reflection over an arbitrary set of (on-disk) assemblies, in order to allow type providers to do correct binding. 

Sooooo.... This PR solves problem 2 by extendending the CommonProviderImplementation with an implementation of the Type/MethodInfo/... reflection objects over arbitrary on-disk assemblies.  It uses a bespoke from-the-bytes assembly reader derived from the Abstract IL code (AssemblyReader.fs) which reads sufficient information for the reflection we need to do.

It also extensively rewrites (and greatly simplifies) the AssemblyReplacer logic of FSharp.Data.  The distinction between Runtime and DesignTime types is now fully tracked by a type distinction between Type/Expr and TypeRT/ExprRT, and the transformations in and out of these types are fully encapsulated in a set of alternative ProvidedTypeDefinition/ProvidedMethodDefinition/... construction functions which apply the transformations at exactly the ProvidedTypes boundary

The PR has so far been tested manually on simple Profile47, Profile259 and Profile47 compilations with Visual F# Tools 3.1 and 4.0. I'm still getting full tests to pass.

The whole idea here is really to prototype an end-to-end solution to multi-targeting to find out what the blockers are, and what the F# compiler might provide in F# vNext, but at the same time to get something that works for F# 3.1 and F# 4.0.

I may factor AssemblyReader.fs and ProvidedBindingContext.fs out into independent paket/nuget files since they may be useful in other projects. Also, once all this is stable I'd like to see it made more generally available alongside ProvidedTypes.fs


ovatsus:
This is great! I won't have time to look at this straight away, but definitely will look in detail as soon as I can

tpetricek:
Wow, thanks a lot for looking into this. I'll try to have a detailed look in a day or two!

pezipink:
Awesome!! I will check this out soon as well

dsyme:
I'm getting rate limitation errors from Travis - any thoughts on what to do about this?

    752  2) FS3033: /Users/travis/build/fsharp/FSharp.Data/tests/TestApps/Test.fs(10,15): The type provider 'ProviderImplementation.JsonProvider' reported an error: Cannot read sample JSON from 'https://api.github.com/repos/fsharp/FSharp.Data/issues': The remote server returned an error: (403) Forbidden.Response from https://api.github.com/repos/fsharp/FSharp.Data/issues:{"message":"API rate limit exceeded for 208.78.110.199. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}

dsyme:
This work is pretty much ready to go in, however I'm having trouble getting both Travis and AppVeyor to pass. AppVeyor is giving me this strange problem where there is a hang from 07:46 to 24:14

```
[00:07:02] 
[00:07:46] ***** FSharp.Data.DesignTime.Tests.DocumentationTests.Documentation generated correctly ("library/XmlProvider.fsx")
[00:07:46] Processing 'library/XmlProvider.fsx
[00:07:46] '
[00:07:46] 
[00:24:14] Running build failed.
[00:24:14] Error:
[00:24:14] System.Exception: Process C:\projects\fsharp-data\packages\NUnit.Runners\tools\nunit-console.exe "-nologo" "-noshadow" "-labels" "C:\projects\fsharp-data\tests\FSharp.Data.DesignTime.Tests\bin\Release\FSharp.Data.DesignTime.Tes
[00:24:14] ts.dll" "-xml:TestResults.xml" "-framework:4.5" "-domain:Multiple"  timed out.
[00:24:14]    at Fake.ProcessHelper.ExecProcessWithLambdas@89-22.Invoke(String message) in C:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 89
```

Travis is giving me this problem which looks like running out of resources in some way:

```
***** FSharp.Data.DesignTime.Tests.SignatureTests.Validate signature didn't change (Html,doctor_who2.html,False,False,)

***** FSharp.Data.DesignTime.Tests.SignatureTests.Validate signature didn't change (Html,SimpleHtmlLists.html,False,False,)

***** FSharp.Data.DesignTime.Tests.SignatureTests.Validate signature didn't change (Html,zoopla.html,False,False,)

***** FSharp.Data.DesignTime.Tests.SignatureTests.Validate signature didn't change (Html,zoopla2.html,False,False,)

create_thread: CreateThread error 0x0

Unhandled Exception:

System.SystemException: Thread creation failed.



Server stack trace:

   at System.Threading.Thread.StartInternal (IPrincipal principal, System.Threading.StackCrawlMark& stackMark) in <filename unknown>:line 0

   at System.Threading.Thread.Start (System.Threading.StackCrawlMark& stackMark) in <filename unknown>:line 0

   at System.Threading.Thread.Start () in <filename unknown>:line 0

   at NUnit.Util.DomainManager+DomainUnloader.Unload () in <filename unknown>:line 0

   at NUnit.Util.DomainManager.Unload (System.AppDomain domain) in <filename unknown>:line 0

   at NUnit.Util.TestDomain.Unload () in <filename unknown>:line 0

   at NUnit.Util.TestDomain.Dispose () in <filename unknown>:line 0

   at NUnit.Util.AggregatingTestRunner.Dispose () in <filename unknown>:line 0

   at NUnit.Core.ProxyTestRunner.Dispose () in <filename unknown>:line 0

   at (wrapper managed-to-native) System.Runtime.Remoting.RemotingServices:InternalExecute (System.Reflection.MethodBase,object,object[],object[]&)

   at System.Runtime.Remoting.RemotingServices.InternalExecuteMessage (System.MarshalByRefObject target, IMethodCallMessage reqMsg) in <filename unknown>:line 0

```

ovatsus:
We had issues before, we need to replace that GitHub usage of JsonProvider for something else in the ConsoleTests. Created issue: https://github.com/fsharp/FSharp.Data/issues/878

tests/FSharp.Data.DesignTime.Tests/SignatureTests.fs
```diff
@@ -32,29 +33,113 @@ let expectedDirectory = sourceDirectory ++ "expected"
...
+    | 47 -> referenceAssembliesPath ++ "Framework" ++ ".NETPortable" ++ "v4.0" ++ "Profile" ++ "Profile47" 
+    | 7 -> referenceAssembliesPath ++ "Framework" ++ ".NETPortable" ++ "v4.5" ++ "Profile" ++ "Profile7" 
+    | 259 -> referenceAssembliesPath ++ "Framework" ++ ".NETPortable" ++ "v4.5" ++ "Profile" ++ "Profile259" 
+    | _ -> failwith "unimplemented portable profile"
+
+let net40FSharp31Refs = [net40AssembliesPath ++ "mscorlib.dll"; net40AssembliesPath ++ "System.Xml.dll"; net40AssembliesPath ++ "System.Core.dll"; net40AssembliesPath ++ "System.Xml.Linq.dll"; net40AssembliesPath ++ "System.dll"; fsharp31AssembliesPath ++ "FSharp.Core.dll"]
+let portable47FSharp31Refs = [portableAssembliesPath 47 ++ "mscorlib.dll"; portableAssembliesPath 47 ++ "System.Xml.Linq.dll"; fsharp31PortableAssembliesPath 47]
+
+let portableCoreFSharp31Refs profile = 
+    [ for ass in [ "System.Runtime"; "mscorlib"; "System.Collections"; "System.Core"; "System"; "System.Globalization"; "System.IO"; "System.Linq"; "System.Linq.Expressions";
```
ovatsus: Maybe use `for asm in` instead? :)

ovatsus:
Somewhat unrelated: currently most of the runtime assembly files are reincluded in the designtime assembly, because of issues finding the runtime dll while the designtime dll is being invoked. Do you think that can be also fixed?

src/CommonProviderImplementation/QuotationBuilder.fs
```diff
@@ -53,25 +53,18 @@ let (?) (typ:Type) (operation:string) (args1:'T) : 'R =
...
           pi.CanWrite && pi.GetSetMethod().IsStatic
-        if isStatic then Expr.PropertyGet(pi, args)
-        else Expr.PropertyGet(List.head args, pi, List.tail args)
+        if isStatic then Expr.PropertyGetUnchecked(pi, args)
+        else Expr.PropertyGetUnchecked(List.head args, pi, List.tail args)
     | options -> failwithf "Constructing call of the '%s' operation failed. Got %A" operation options
 
-  // If the result is a function, we are called with two tuples as arguments
-  // and the first tuple represents type arguments for a generic function...
-  if FSharpType.IsFunction(typeof<'R>) then
```
ovatsus: Why isn't this needed anymore?

src/CommonProviderImplementation/QuotationBuilder.fs
```diff
@@ -53,25 +53,18 @@ let (?) (typ:Type) (operation:string) (args1:'T) : 'R =
     | [| :? MethodInfo as mi |] -> 
         let mi = 
           if tyargs = [] then mi
-          else mi.MakeGenericMethod(tyargs |> Array.ofList |> Array.map ProvidedTypeDefinition.EraseType)
```
ovatsus: Why is eraseType no longer needed? IIRC it was needed when tyargs contained units of measure

src/CommonProviderImplementation/Helpers.fs
```diff
@@ -147,43 +148,18 @@ module ProviderHelpers =
           SampleIsWebUri : bool
           SampleIsResource : bool }
 
-    type private EmbeddedResourceReader() =
```
ovatsus: This AppDomain marshalling was needed to not lock the files, otherwise it was very hard to build FSharp.Data.Toolbox.Twitter. Has something else changed that makes this not needed any more?

src/CommonProviderImplementation/AssemblyReplacer.fs
```diff
@@ -14,145 +15,128 @@ open Microsoft.FSharp.Quotations.Patterns
...
-          if newT = null then
-            None
-          else
-            typeCache.Add((asm, t), newT)
-            Some newT
-    match List.choose getType asmCandidates with
-    | [ t ] -> t
-    | _ -> failwithf "Type '%O' not found in %A" t (asmCandidates |> List.map (fun asm -> asm.Location))
+            match Array.choose (tryGetTypeFromAssembly fullName) asms |> Seq.distinct |> Seq.toArray with
+            //| [| (newT, canSave) |] ->
```
ovatsus: Did you meant to remove this commented out section or is it work in progress?

ovatsus:
Would you mind also updating Test.fsx? I usually rename it to Test.fs and change the designTime project to console to run and debug the generation code. It's missing the assemblyRefs param

src/WorldBank/WorldBankProvider.fs
```diff
@@ -20,7 +20,7 @@ open FSharp.Data.Runtime.WorldBank
 type public WorldBankProvider(cfg:TypeProviderConfig) as this = 
     inherit DisposableTypeProviderForNamespaces()
 
-    let asm, version, replacer = AssemblyResolver.init cfg
+    let bindingCtxt, asm, version, replacer = AssemblyResolver.init cfg // TODO: make this respect the binding context
```
ovatsus: What is missing here?

src/CommonProviderImplementation/AssemblyResolver.fs
```diff
@@ -7,211 +11,54 @@ open System.Reflection
...
-    let runtimeAssembly = 
-        if useReflectionOnly then Assembly.ReflectionOnlyLoadFrom cfg.RuntimeAssembly
-        else Assembly.LoadFrom cfg.RuntimeAssembly
+        match fsCore with 
+        | Choice2Of2 _err -> FSharpDataRuntimeInfo.Net40
+        | Choice1Of2 asm -> 
+           let ver = asm.GetName().Version
+           match ver.Major, ver.Minor with 
+            | 4, _ -> FSharpDataRuntimeInfo.Net40 // 4.3.0.0, 4.3.1.0, 4.4.0.0
+            | _ -> FSharpDataRuntimeInfo.Portable_47_7_259
```
ovatsus: Could this assumption break in the future and we get a portable profile with major version 4?

ovatsus:
Since we now have these untyped quotations, maybe we can now create FSharpFunc's from other framework types, and remove the workaround of making all function types delegate types?

src/CommonProviderImplementation/DebugProvidedTypes-modified.fs
```diff
@@ -0,0 +1,556 @@
...
+                | _ when fullName = "System.Boolean" -> "bool"
+                | _ when fullName = "System.Object" -> "obj"
+                | _ when fullName = "System.Int32" -> "int"
+                | _ when fullName = "System.Int64" -> "int64"
+                | _ when fullName = "System.Double" -> "float"
+                | _ when fullName = "System.Single"-> "float32"
+                | _ when fullName = "System.Decimal" -> "decimal"
+                | _ when fullName = "System.String" -> "string"
+                | _ when fullName = "System.Void" -> "()"
+                | _ when fullName = typeof<unit>.FullName -> "()"
```
ovatsus: Aren't these two cases the same?

ovatsus:
GitHub doesn't show the diff of all files, so can't leave comments inline

ProvidedTypes-modified.fs line 1573:
> // TODO: this is not a correct implementation.  It would be better to have FSharp.Data pass in DeclaredOnly.

Could you please expand a bit more on how to fix this? IIRC this GetMembers is not called directly so how can we fix it?

ovatsus:
I won't pretend to understand all the black magic going on here, but overall LGTM. Finally no more worry about those toRuntime/toDesignTime! There's probably stuff that can be simplified because it was trying to compensate for that need
I'll leave it here a couple more days for Thomas and Ross to look and I'll do some testing on WinRT with profile 7 and 259 where profile 47 didn't work correctly

I'd also maybe only keep full .NET 4.0 and profile 259, as 259 works in all places where 47 and 7 would work. What do you think?

tests/FSharp.Data.DesignTime.Tests/SignatureTests.fs
dsyme: noted, thanks

src/CommonProviderImplementation/QuotationBuilder.fs
dsyme: It turned out this case was never used in FSharp.Data, I simplified it to match.

src/CommonProviderImplementation/Helpers.fs
dsyme: If we're reading resources via AssemblyReader, then that is using an in-memory snapshot of the bytes of the target assembly, and so there should be no file locking.

(Mind you, that's also chewing through a non-trivial amount of memory, which needs to be solved, I'm trying to work out what we can do about that - if anything)

src/CommonProviderImplementation/QuotationBuilder.fs
ovatsus: I remember fighting with this bit of code, so it was used before. Maybe it was part of Apiary...

src/CommonProviderImplementation/Helpers.fs
ovatsus: Ah, right, we're no longer calling Assembly.Load, so it won't be locked

src/CommonProviderImplementation/AssemblyReplacer.fs
dsyme: The current problem here is multiple references to the same assembly, which result in multiple (logically identical) type bindings.  In theory we don't get multiple references to the same assembly, but I was wondering if we should just tolerate them and take the first binding.

src/WorldBank/WorldBankProvider.fs
dsyme: Nothing, old comment

src/CommonProviderImplementation/AssemblyResolver.fs
dsyme: Yes.  AFAICS the version info is only used for determining if a local file system is supported by the target runtime, I wonder if we can replace it with something more robust

src/CommonProviderImplementation/AssemblyReplacer.fs
ovatsus: Humm, could it happen that FSharp.Core 4.3.0.0 is referenced because that's what FSharp.Data is dependent on and FSharp.Core 4.3.1.0 or 4.4.0.0 is also referenced because that's the version currently loaded in the TP host inside the compiler? I think I've seen something like that before

src/CommonProviderImplementation/DebugProvidedTypes-modified.fs
dsyme: One is unit, one is void (nb. essentially the same as the printing code that was already there).

src/CommonProviderImplementation/AssemblyResolver.fs
ovatsus: Maybe check if the type File is defined?

dsyme:
> Somewhat unrelated: currently most of the runtime assembly files are reincluded in the designtime assembly, because of issues finding the runtime dll while the designtime dll is being invoked. Do you think that can be also fixed?

At the logical level I quite like the current design: it makes for a very clean mapping between design time types and runtime types, no matter what the target framework, and keeps the design time type provider component a single DLL.


> Since we now have these untyped quotations, maybe we can now create FSharpFunc's from other framework types, and remove the workaround of making all function types delegate types?

Yes, I think so.

> Could you please expand a bit more on how to fix this? IIRC this GetMembers is not called directly so how can we fix it?

IIRC GetMembers is only called during printing by the debug dumping code.  I think now that code is passing in DeclaredOnly (which is sufficient) then we should simply remove the ability to do GetMembers up the inheritance chain (failing that case instead)

dsyme:
> I'd also maybe only keep full .NET 4.0 and profile 259, as 259 works in all places where 47 and 7 would work. What do you think?

One problem is that profile 259 doesn't define "OptionalAttribute", which the F# compiler uses to create the ``[opt]`` IL metadata used to indicate an optional parameter to C#.  So C# clients of the Profile 259 DLL don't see optional arguments on methods which is a bit unfortunate.

BTW one of the reasons I'm doing all this is that AFAIK the x-platform .NET Core target will effectively be a portable-like profile target compatible with .NET 4.x.  So in the future more type providers will need to target non .NET-4.x profiles.

dsyme:
@ovatsus - see this question: https://github.com/fsharp/FSharp.Data/commit/aed22f01144f16a5b689c0408e8af8dbd4bd2163#commitcomment-13458945

Repeated below to keep the discussion in this thread:

> How critical is this cache? It seems structurally wrong to cache provided types in a global and retrieve them from later type providers - logically speaking this in turn holds handles to all referenced assemblies (these assemblies were previously reflection assemblies, but #875 replaces them with AssemblyReader-backed assemblies. These consume memory and should really only need to live as long as the ITypeProvider instance to which they are related - but this cache means they live on longer).

ovatsus:
Unfortunately without this cache, IntelliSense gets very sluggish, because of some inefficiencies in the compiler that cause the TP to be instantiated multiple times. See the this comment: https://github.com/fsharp/FSharp.Data/issues/300#issuecomment-37104040

dsyme:
I see, thanks.  So it's particularly in the incremental scenario where F# does complete invalidation, and you're try to avoid recomputing everything.  That's fair enough.

dsyme:
@ovatsus 

AppVeyor is passing at last, good to see.

- I kept the cache, though not when running inside FSI.EXE (where no validation happens). This helps the documentation generation tests to go through.

- I reduced memory usage in a bunch of somewhat obvious ways (through weakly caching assembly readers and some changes in AssemblyReader representations).  There are more savings to be made if needed. The main memory costs are the backing ``byte[]`` for the in-memory copies of the referenced DLLs.

It looks like design-time memory usage of the FSharp.Data type provider will go up with this PR - that's hard to avoid. However I believe it is now closer to acceptable. At least the fairly large stress test of generating documentation now passes, and opening FSharp.Data.Tests solution and looking through all the files stays in good memory bounds.  

I'm still concerned by the strong cache mentioned above - there is a potential for that to consume a considerable amount of memory (more than before due to using our own assembly reading) which could be worse than avoiding recompute. Let's think that over.

ovatsus:
I can try to disable the cache and measure the impact again. Ideally, this would be improved on the compiler aide so TPs aren't invoked so often.

This just occurred to me, since we no longer load the assemblies, maybe now the dll locking will be better?

dsyme:
@ovatsus I can see that the caches change an incremental invalidation/update response from O(n) to O(1) where n is the number of type provider instantiations in the script/project.  So don't remove the cache - it's definitely needed in some form - we'll just need to think how to make it less liable to leak and/or use resources for long periods.

DLL locking problems should be generally improved, though the design-time assembly itself will remain locked as always with Visual Studio.

dsyme:
Yay, appveyor and travis both passing now.  (I had to disable GitHub in the documentation though)

dsyme:
@ovatsus  Random thought: what about if we only write the entries to the invalidation-recovery cache just when then tp invalidation occurs? e.g. change:

            providedTypesCache.[key] <- CacheValue (providedType, fullKey, DateTime.Now)
            tp.AddDisposeAction fullTypeName <| fun () -> 
                log (sprintf "Invalidating cache for %s [%d]" fullTypeName tp.Id )
                providedTypesCache.Remove key |> ignore

to something like this:

            tp.AddDisposeAction <| fun typeNameBeingDisposed -> 
                if typeName <> typeNameBeingDisposed then 
                    providedTypesCache.[key] <- CacheValue (providedType, fullKey, DateTime.Now)

meaning we save all the types except the one being disposed.  This would mean the captured resources are scoped to the tp object (in the situation where no invalidation is occurring), and the global variable cache need only hold references for the duration of an invalidation + recovery

dsyme:
Update - the approach of only saving types temporarily on incremental disposal does seem to work, I've added that to the PR though we could separate that out.

ovatsus:
I think this is mostly ready to merge now

dmitry-a-morozov:
@dsyme Great work. 
Let me know when it's stable - I will try apply to FSharp.Data.SqlClient

dsyme:
@dmitry-a-morozov Is FSharp.Data.SqlClient generative?  I think that will require more work. 

I'm thinking that the next step is to move the various independent pieces of this across to the type provider starter pack and adding specific independent tests for both non-generative and generative cases.

dsyme:
@ovatsus Yes, fix for #870 included.

ovatsus:
Merged, we need to update the readme and docs and maybe relase this as a beta version

dmitry-a-morozov:
@dsyme Both SqlCommandProvider and SqlProgrammabilityProvider are erased types. Only SqlEnumProvider is generative.

dsyme:
@dmitry-a-morozov OK, thanks.

dsyme:
I recorded the problems with F# quotations that led to UncheckedQuotations.fs here: http://fslang.uservoice.com/forums/245727-f-language/suggestions/9980001-make-f-quotations-and-fsharp-reflection-usable-wi

The workaround here is adequate but ideally we should do it right

PR #883, opened 2015-10-11
Fixed parsing of tables with multirow headers (closes #877)

### LINKED ISSUE #877: Header being considered as data row in HTMLProvider
https://en.wikipedia.org/wiki/List_of_Doctor_Who_serials

    let doctorWho = new HtmlProvider<"http://en.wikipedia.org/wiki/List_of_Doctor_Who_serials">()
    let row = doctorWho.Tables.Overview.Rows.[0]
    val row : HtmlProvider<...>.Overview.Row =  ("", "Season/Series", "Doctor(s)", "Serials", "Episodes", "First aired",   "Last aired", "Premiere", "Finale", "Average")

ovatsus:


src/Html/HtmlRuntime.fs
```diff
@@ -183,16 +191,33 @@ module HtmlRuntime =
...
+            | 1 -> true, res
+            | _ ->
+                for i = 1 to numberOfHeaderRows - 1 do
+                    for j = 0 to numberOfColumns - 1 do
+                        let previousCell = res.[i-1].[j]
+                        let thisCell = res.[i].[j]
+                        if previousCell.Data <> "" && thisCell.Data <> "" && thisCell.Data <> previousCell.Data then
+                            res.[i].[j] <- Cell(true, previousCell.Data + " - " + thisCell.Data)
+        
+                true, res.[numberOfHeaderRows-1..]
```
ovatsus: I would prefer to leave the table structure intact and just do this at the TP level, but it was getting quite complicated, so I left it like this. @colinbull any better suggestion?

colinbull: This looks fine, Although I'm surprised, this wasn't handled by this https://github.com/fsharp/FSharp.Data/blob/master/src/Html/HtmlRuntime.fs#L152. Am I missing something?

ovatsus: No it's not. TBH, I didn't understand very well what that bit is doing

PR #886, opened 2015-10-20
Cssselectors

rflechner:
Hi,

Adds CssSelectorExtensions on HTML parser.

This proposed feature is based on the JQuery selectors.
I think that usage of CSS selectors is a very natural way to parse HTML when we are familiarized with Web developments.

A part of selectors described here https://api.jquery.com/category/selectors/ are implemented.
It's possible to implement the missing selectors if you are interested by this feature.

I wrote tutorials and documentations in FSharp.Data\docs\content\library\HtmlCssSelectors.fsx (compliant with the documentation generator).

Implementation is in FSharp.Data\src\Html\HtmlCssSelectors.fs
Tests are in FSharp.Data.Tests\HtmlCssSelectors.fs

Regards.

Romain.


docs/content/library/HtmlCssSelectors.fsx
```diff
@@ -0,0 +1,308 @@
...
+*)
+let doc = HtmlDocument.Load "http://www.google.co.uk/search?q=FSharp.Data"
+
+(**
+To be sure we get search results, we will parse links in the div with id `search`.
+Then, for example, we could ensure we the HTML's structure is really compliant with the parser
+using the direct descendants selector.
+*)
+
+let links = doc.CssSelect "div#search > div#ires li.g > div.s div.kv cite"
```
tpetricek: Could you change this so that the spacing looks as follows: 

    let links = 
      doc.CssSelect "div#search ..."
      |> List.map (fun n ->
          match n.InnerText() with
          | ... )

This makes the code sample a bit less wide and it will look better in the generated HTML page.

src/Html/HtmlCssSelectors.fs
```diff
@@ -0,0 +1,375 @@
+﻿namespace FSharp.Data
+
+open System
+open FSharp.Data
+open System.Runtime.CompilerServices
+
+module HtmlCssSelectors =
```
tpetricek: Is this a part of the public interface? If so, it would be good to document it using `///`.

But as far as I can see, this is only internals that we don't need to expose. Could you mark it as `module internal HtmlCssSelectors = ...` ?

src/Html/HtmlCssSelectors.fs
```diff
@@ -0,0 +1,375 @@
...
+                    //failwith "Invalid css selector syntax"
+                    []
+            tokenize' [] source
+
+    type FilterLevel = 
+        | Root
+        | Children
+        | Descendants
+
+    type CssSelectorExecutor(nodes:HtmlNode list, tokens:SelectorToken list) =
```
tpetricek: Would it make sense to convert this from a class into a function? (And possibly extract the nested functions that are not accessing `nodes` and `tokens` and does not need to be nested?)

(I do not mind too much - just wondering if a class is needed here - even if function would work, I'm quite happy to keep this as it is - it looks fine to me!)

tpetricek:
This is a fantastic piece of work! I'll let @ovatsus and @colinbull comment too, but I think we should definitely merge this.

I added a couple of small comments - if you can have a look at them, that would be great. But it's nothing major - it looks great as it is.

rflechner:
Thanks a lot for the code review.
I will try to apply your suggestions to src/Html/HtmlCssSelectors.fs

colinbull:
This is absolutely awesome, it has always bugged me that we haven't had anything like this. My only comment is it would be nice to enumerate the implemented / missing selectors somehow in the documentation. Just so people aren't surprised when things don't work; apart from that nice work :+1:

tpetricek:
+1 to what @colinbull says. Having a brief table would be neat :-).

You should be able to create tables in the docs using the syntax from here: https://github.com/tpetricek/FSharp.Formatting/pull/9

rflechner:
Yes, good idea.
It will be easier to choose selectors that I must implement.

ovatsus:
This looks great! I'll have a more detailed look over the weekend

src/Html/HtmlCssSelectors.fs
```diff
@@ -0,0 +1,375 @@
...
+        let getOffset (t:List<char>) = 
+                charCount - 1 - t.Length
+
+        member public x.Tokenize(pCssSelector:string) =
+            cssSelector <- pCssSelector
+            source <- Array.toList(cssSelector.ToCharArray())
+            charCount <- source.Length
+            x.tokenize()
+            
+        member private x.tokenize() =
```
thinkbeforecoding: For parsing speed, I tend to use a struct like
```
[<Struct>]
type State(source: string, pos: string) =
    member __.Source = source
    member __.Pos = pos
```
and a module with some inline functions, like next etc.. And an inline active record to match current char and tail..
You don't get the :: syntax so it leads to more parenthesis nesting.. but it works quite well.

Since css selector are usually short, I  don't know if it'll make much difference anyway..

src/Html/HtmlCssSelectors.fs
```diff
@@ -0,0 +1,375 @@
...
+                    tokenize' (EmptyNode(getOffset t + 1) :: acc) t
+                | TokenStr ":disabled" t ->
+                    let _, t' = readString "" t
+                    tokenize' (Disabled(getOffset t + 1) :: acc) t'
+                | StartsWith ":enabled" t ->
+                    tokenize' (Enabled(getOffset t + 1) :: acc) t
+                | StartsWith ":file" t ->
+                    tokenize' (File(getOffset t + 1) :: acc) t
+                | '>' :: t ->
+                    let seqtoken = acc |> List.toSeq |> Seq.skip(1) |> Seq.toList
```
thinkbeforecoding: You can use acc |> List.tail directly here

src/Html/HtmlCssSelectors.fs
```diff
@@ -0,0 +1,375 @@
...
+                    | x, [] -> Some(x)
+                    | xh :: xt, sh :: st when xh = sh -> equal xt st
+                    | _ -> None
+
+                equal x chars
+
+            let rec tokenize' acc sourceChars = 
+                match sourceChars with
+                | w :: t when Char.IsWhiteSpace(w) -> 
+                    let seqtoken = acc |> List.toSeq |> Seq.skip 1 |> Seq.toList
```
thinkbeforecoding: You can use acc |> LIst.tail here

src/Html/HtmlCssSelectors.fs
cboudereau: Good job! This is the future in order to replace htmltable and a more
flexible way to parse Html!

src/Html/HtmlCssSelectors.fs
```diff
@@ -0,0 +1,375 @@
...
+                    []
+            tokenize' [] source
+
+    type FilterLevel = 
+        | Root
+        | Children
+        | Descendants
+
+    type CssSelectorExecutor(nodes:HtmlNode list, tokens:SelectorToken list) = 
+        let mutable level = FilterLevel.Descendants
```
thinkbeforecoding: You could pass  ```level```  in recursion, and use it as an argument to ```getTargets``` and other functions.
This would make th intent of level simpller and probably remove the need for a class.

thinkbeforecoding:
This seems great !
And it would be awesome to have function type provider for this ! Checking at compile time that the CssSelector is valid and returns some results based on examples !

rflechner:
Hi,

Do I need to change anything else or the pull request is ok for a merge ?

Regards.

ovatsus:
This overall looks good, but as it's a major new feature I'd like to review it carefully, but free time is not helping out, sorry about that. I'll try to have a better look next weekend

rflechner:
Hi,

As asked on twitter (https://twitter.com/action_dk/status/684618813975875584),
waiting for the merge, I created a small repo (with ProjectScaffold) containing some scraping tools and I integrated the CSS selectors.
http://rflechner.github.io/ScrapyFSharp/HtmlCssSelectorsExample.html

I just got a small problem when running unit tests, 
because it seems the runner cannot find extension methods declared in HtmlOperations


Thanks for your enthousiasm,

Romain.

tpetricek:
Sorry for the massive delay here. I'm just looking at the PR and will report back soon. It would be lovely to get this merged!

tpetricek:
I did a couple of changes to address the above feedback. The final version is in #932. I think that's good to merge now.

rflechner:
great!
thanks

ovatsus:
Finally merged. I'll also merge a couple of other things pending and release 2.3.0-beta1 tomorrow with this

PR #894, opened 2015-12-06
Improve parsing performance

eulerfx:
...by pre-allocating StringBuilder, and ResizeArray for records and arrays. Also, reduce allocations and ops during numeric parsing by avoiding use of Option and active patterns.

This resulted in about a 20% parsing performance improvement in terms of time, as well as reduced gen 0 and gen 1 collections:

> Real: 00:00:25.131, CPU: 00:00:25.109, GC gen0: 3445, gen1: 4, gen2: 1

vs 

> Real: 00:00:19.631, CPU: 00:00:19.640, GC gen0: 1689, gen1: 3, gen2: 1

The test parsed a 4KB JSON string 500k times.

For comparison, parsing using `Newtonsoft.Json.Linq.JToken.Parse` recorded:

> Real: 00:00:38.252, CPU: 00:00:38.234, GC gen0: 3363, gen1: 305, gen2: 1

### DIFF
--- a/src/Json/JsonValue.fs
+++ b/src/Json/JsonValue.fs
@@ -149,10 +149,8 @@ type private JsonParser(jsonText:string, cultureInfo, tolerateErrors) =
     let mutable i = 0
     let s = jsonText
 
-    // pre-allocate buffers for strings, arrays and records
-    let buf = StringBuilder()
-    let pairs = ResizeArray<_>()
-    let vals = ResizeArray<_>()
+    
+    let buf = StringBuilder() // pre-allocate buffers for strings
 
     // Helper functions
     let skipWhitespace() =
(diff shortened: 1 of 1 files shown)


eulerfx:
I also couldn't build the test project as is. I ran Json tests only by excluding provider modules.

ovatsus:
Didn't know we were already faster than JSON.NET :)

eulerfx:
@ovatsus ahh nice catch! Luckily, the array and pairs perf improvement was marginal - more-so the allocation reduction. The StringBuilder was the biggie. The numeric parsing was also marginal, maybe just under a second improvement. I'm assuming due to option. Trying to parse int64 first also showed improvement, but it is obviously dependent on the input. I'll try it again, and update the PR.

Another win will be from having numeric parsers which take string ranges. We can either wait for a BCL update are internalize the substring parsing implementations.

eulerfx:
Fixed invalid use of buffers and reverted numeric parsing. Still very good!

> Real: 00:00:19.727, CPU: 00:00:19.734, GC gen0: 1888, gen1: 4, gen

ovatsus:
Thanks!

Issue #901, opened 2015-12-26
Cookies with "http://"-prefixed domains cause exception


taylorwood:
I've been using the `Http.Request` methods to hit various sites, and I've run into some edge cases where a site includes the protocol on the domain in a cookie: `domain=http://site.com`. This causes `System.Net.Cookie` to throw `The 'Domain'='http://site.com' part of the cookie is invalid.` Browsers accept these (invalid?) cookies without issue.

I saw there's already a function for parsing multiple cookies from one `Set-Cookie` header, so I thought it may be acceptable to handle this case as well.

PR #902, opened 2015-12-26
Handle cookies with "http://"-prefixed domain value.

taylorwood:
Proposed fix for issue #901.

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -680,11 +680,12 @@ module private HttpHelpers =
                         cookie.Path <- kvp.[1]
                 elif cookiePart.IndexOf("domain", StringComparison.OrdinalIgnoreCase) = 0 then
                     let kvp = cookiePart.Split '='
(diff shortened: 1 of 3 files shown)


src/Net/Http.fs
```diff
@@ -680,8 +680,13 @@ module private HttpHelpers =
                         cookie.Path <- kvp.[1]
                 elif cookiePart.IndexOf("domain", StringComparison.OrdinalIgnoreCase) = 0 then
                     let kvp = cookiePart.Split '='
-                    if kvp.[1] <> "" then
-                        cookie.Domain <- kvp.[1]
+                    let domain = // handle errant protocol in domain
+                        if kvp.[1].StartsWith("http://", StringComparison.OrdinalIgnoreCase) then
```
ovatsus: Maybe we should do the same for https?

taylorwood: Yeah, pushed another commit.

ovatsus:
Thanks

PR #903, opened 2015-12-29
Fixed the never expiring in-memory cache

bsulima:
Cache timestamp should be deducted from the current time, not the other
way round


ovatsus:
Thanks

Thorium:
The cache is not actually never cleared. So this will lead to out-of-memory in a long run.

PR #905, opened 2016-01-17
HTTP error response stream FIX

draganjovanovic1:
HTTP error response stream preserved in the augmentWebExceptionsWithDetails function


tpetricek:
First of all, sorry for the massive delay!

I like the idea, but I'm not sure how is this supposed to work. Can you please add a brief comment explaining what is the problem solved here? (Or if this refers to an issue, can you add a link?)

draganjovanovic1:
It does not refer to an existing issue, it is something I experienced when used the library. I will try to describe it in more details.

I had to read a response stream when HTTP call fails (status code > 399) but I wasn't able to do so as stream was being read internally by the lib (lines 443-449) to format WebException text. Line 448 didn't work as response stream is not seekable (that's why an empty catch is there probably) and I was getting an exception when trying to read the stream. The solution was to copy response stream to an instance of MemoryStream to be able to read it more than once. WebResponse decorator is needed only to pass a new response stream to a WebException instance.

tpetricek:
Thanks, that makes sense! So the exception happens at some point during the reading of the stream and the decorator lets you re-read the part that has been read before the error?

draganjovanovic1:
Yes. An exception occurs when you try to read a response stream. With the decorator a stream is still being read by the lib to format WebException text and is also available to caller to read it again.

tpetricek:
Hmm... so if you have "hello world" (in theory), but the download crashes after "hello", you'll have a memory stream that contains "hello" and you can read it again?

draganjovanovic1:
Actually no, I think we misunderstood each other. 

The problem is that I the stream is not readable at all as it was already read by lib internally. It wasn't failure during the read.

tpetricek:
Hmm... I might still be a bit lost here. Sorry about that :-).

I can see that the stream can be read just once. But that means:

 1. Either we read it - but then no exception happens
 2. We do not even get to the point where we start reading it (but then we can wrap it in the exception and the client can read it, because we did not touch it)
 3. We start reading it, but then exception happens - what we then wrap in the exception is useless, because you cannot seek to the beginning.

I thought your PR is addressing the case (3), because it would not be a problem for (1) or (2). Am I missing another scenario?

draganjovanovic1:
Just to clarify, there is no issue with reading response stream when HTTP status code is OK (< 400).

The problem is when HTTP call returns an error (400 and above). In that case response stream is not readable by a caller as library is reading it to format WebException text. Here is the code doing that:

```fsharp
let responseExn =
  try
    use responseStream = exn.Response.GetResponseStream()
    use streamReader = new StreamReader(responseStream)
    let response = streamReader.ReadToEnd()
    try 
      // on some platforms this fails
      responseStream.Position <- 0L
    with _ -> ()
    if String.IsNullOrEmpty response then None
    else Some(WebException(sprintf "%s\nResponse from %s:\n%s" exn.Message exn.Response.ResponseUri.OriginalString response, exn, exn.Status, exn.Response))
with _ -> None
```

As you can see there is an attempt to rewind response stream (responseStream.Position <- 0L) but that won't work as stream is not seekable. This is the part which is fixed by reading response stream to a MemoryStream first to enable multiple reads. Then we read it once to populate WebException text and rewind it to give it to a caller as it is an original.

I hope I was clearer now :-)

ovatsus:
The reason we read the responseStream on errors, is to be able to give meaningfull error messages on FSI, which doesn't happen when using WebRequest or HttpClient directly

ovatsus:
Tried to merge this, but it only works on .NET desktop, not on the portable profiles, could you please fix it and then I'll merge?

tpetricek:
Okay, I think I understand this now :-).

If it is just for error reporting, wouldn't it make sense to just return `response` that we already read from the response stream as part of the exception? (I guess we'd need our own exception type for that, but it might be simpler...)

ovatsus:
That's what being done currently, but the problem is that the response is read once

tpetricek:
Oh, I don't mean the response stream, but the actual `responseText` string we read.. (but that's assuming it fails after we read it... - so not the case (3) in my above comment.

draganjovanovic1:
The problem with portable profiles is that WebResponse portable class does not have IsFromCache and IsMutuallyAuthenticated properties and Close method. I can comment those overrides to make it work for all profiles but I would like to keep overrides for .NET desktop.

Should I add new compiler symbols to achieve this like you already did for e.g. FX_NO_WEBREQUEST_CONTENTLENGTH, or you have some other idea?

draganjovanovic1:
@ovatsus I have pushed fix for portable profiles, please review.

ovatsus:
It seems to still be failing on AppVeyor:



    "C:\projects\fsharp-data\FSharp.Data.sln" (Rebuild target) (1) ->
    "C:\projects\fsharp-data\src\FSharp.Data.fsproj" (Rebuild target) (2) ->
    (CoreCompile target) -> 
      C:\projects\fsharp-data\src\Net\Http.fs(355,20): error FS0859: No abstract property was found that corresponds to this override [C:\projects\fsharp-data\src\FSharp.Data.fsproj]
      C:\projects\fsharp-data\src\Net\Http.fs(355,42): error FS0039: The field, constructor or member 'SupportsHeaders' is not defined [C:\projects\fsharp-data\src\FSharp.Data.fsproj]
      C:\projects\fsharp-data\src\Net\Http.fs(374,39): error FS0039: The field, constructor or member 'Dispose' is not defined [C:\projects\fsharp-data\src\FSharp.Data.fsproj]
     
        0 Warning(s)
        3 Error(s)

draganjovanovic1:
@ovatsus The build was failing on Windows, I didn't try it :) It is fixed now. Please review.

PR #915, opened 2016-03-02
Fields containing double quotes should be enclosed in double-quotes

Karamell:
according to RFC 4180:

> Fields containing line breaks (CRLF), double quotes, and commas
>        should be enclosed in double-quotes.


ovatsus:
Thanks

Issue #802, opened 2015-03-17
How to change the time-out value of request in HTTP Utilities


darting:
Hi all, 

how can I change the time-out value of request? I can not find any documentation. 

thank you very much

colinbull:
Do you mean on a `Http.AsyncRequest` variant or on one of the type providers?

If it is on the `Http.AsyncRequest` variants then there is an optional parameter `customizeHttpRequest` that takes a function in which you can set the timeout, for example, 

      Http.AsyncRequest("http://www.google.com", 
                                      ?customiseHttpRequest = (fun req -> req.Timeout <- 1000; req)

If it is on a the type provider then it is not exposed in the static parameters. However you can do something like the following, 

     type Data = CsvProvider<"sample.csv">

     let data  = Data.Load(Http.RequestStream("http://some/uri", 
                                          ?customiseHttpRequest = (fun req -> req.Timeout <- 1000; req)
                                          ))

Hope that helps

darting:
thank you very mush!!!

ovatsus:
This will need custom logic to implement. Marking as up for grabs

scrwtp:
I've set up a PR after this issue cropped up on SO yesterday.

eriawan:
@ovatsus I think this issue should be closed.
It's been fixed by PR #920  from @scrwtp 🙂

PR #920, opened 2016-03-22
Use HttpWebRequest timeout for async getResponse in HTTP utils

scrwtp:
Solves issue #802 

Note that if caller doesn't provide a timeout through customiseHttpRequest, the default value from HttpWebRequest will still be used (100s). I'd say that follows the principle of least astonishment.


src/Net/Http.fs
```diff
@@ -554,20 +554,39 @@ module private HttpHelpers =
...
-                    | :? WebException as exn -> 
-                        if exn.Response <> null then 
-                           return exn.Response
-                        else 
-                            reraisePreserveStackTrace exn
-                            return Unchecked.defaultof<_>
+                #else
+                    let! child = 
+                        Async.FromBeginEnd(req.BeginGetResponse, req.EndGetResponse)
+                        |> fun x -> Async.StartChild(x, req.Timeout)
```
tpetricek: Could you please split this into two `let` bindings? I know people who prefer piping into lambda functions, but I find it confusing :-). I means something like:

    let op = Async.FromBeginEnd(...)
    let! child = Async.StartChild(...)

tpetricek: Also, would it be possible to somehow detect whether `req.Timeout` has been set and do the custom logic only when it is set? (And call the operation in a normal synchronous way otherwise.) I guess it could have some small overhead associated - not a big deal if this is hard to do though.

tpetricek:
Sorry for the delay here. I added two small comments - if you could address those, that would be fantastic. Otherwise it looks good to me!

src/Net/Http.fs
scrwtp: I don't think detecting `req.Timeout` is an option here - it's a has a meaningful default value when not set, so you can't really differentiate based on that. And the thing with "normal synchronous way" here is that it's really the "async way" being run synchronously. This is different to plain old `HttpWebRequest.GetResponse` in that it doesn't take timeout into account, default or not. 

What I can do instead is scrap the idea of using `HttpWebRequest.Timeout` and instead have an optional timeout argument to the `Http` static members at the end of the file. I suppose making the timeout explicit would help more people than it would confuse.

tpetricek: I see. I do like the idea of having optional `timeout` parameter - that sounds like something that would be quite useful for many people anyway!

That said, if the value is the default, then it probably should be `Timeout.Infinite`, so we could just check for that and use async only when it is not infinite - that should work too.

I'm happy with either of the options.

scrwtp: The "default" default is 100s, but you're right in that we can override it with `Timeout.Infinite` at the time the request is created. For some reason I thought these functions should work with any provided `HttpWebRequest`, but it doesn't seem to be the case and we're creating it internally.

I'll take a look at doing these changes at some point next week.

scrwtp:
This is now done.

ovatsus:
Thanks

PR #932, opened 2016-04-27
Finishing CSS selectors PR (#886)

tpetricek:
This is adding a couple of things on top of #886 before merging it.


ovatsus:
Since this is a major new feature, I'd say do a non-beta release now after the starterpack update and the other small fixes, then after that immediately merge this and do a beta release

ovatsus:
The first documentation example doesn't work:
    let links = 
      doc.CssSelect("div#search > div#ires li.g > div.s div.kv cite")
returns an empty list

ovatsus:
Fixed it, will send another PR with a couple of fixes and merge that one

tpetricek:
I though the bug is probably because the page format has changed - using live source is fragile, but it is probably nicer than having some boring sample page.

PR #933, opened 2016-04-28
Pre and Code HTML tags should keep their content formatted

pirrmann:


tpetricek:
This looks great to me :) I'm merging it - we should have a new release soon-ish.

PR #938, opened 2016-05-01
Allow to get only direct inner text from an HtmlNode (closes #861)

### LINKED ISSUE #861: How to get only inner text from an HtmlNode?
I have such HTML
```HTML
<div class="matchheader">
    <div class="whenm">
        21 minutes ago
        <span style="color: rgb(114, 163, 38); font-weight: bold; text-shadow: 1px 1px 0px #4a7010;"> LIVE</span>
        <span style="color: rgb(209, 33, 33); font-weight: bold;">&nbsp;&nbsp;</span>
    </div>
    <div class="eventm">Do Gaming League</div>
</div>
```
And I want to extract ```21 minutes ago``` from it.
But ```InnerText()``` returns ```21 minutes ago LIVEÂ Â "``` instead.

How to eliminate span elements from inner text?

ovatsus:


ovatsus:
@tpetricek, @colinbull, @xperiandri, any suggestion for a better name than DirectInnerText()?

tpetricek:
I cannot think of better name - the only idea I had was `ImmediateInnerText`, but that's longer and not nicer. So I'd say `DirectInnerText` is good.

PR #950, opened 2016-06-14
Safer assembly replacement - fix CSV provider in FsLab

tpetricek:
More about this here: https://github.com/fsprojects/FSharp.TypeProviders.StarterPack/pull/104

This is a fairly serious bug that means CSV provider does not work when used via FsLab. 

@ovatsus What are your current release plans? Do you think we could release this soon? (Sometime this week would be ideal.)

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -35,9 +35,9 @@ GITHUB
       Octokit
   remote: fsprojects/FSharp.TypeProviders.StarterPack
   specs:
-    src/AssemblyReader.fs (9f0c159defa1d3d9cf554c44acd55fe5b231fb12)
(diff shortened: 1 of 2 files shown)


tpetricek:
(Also, I guess merging #920 as part of the next release makes sense.)

ovatsus:
This is failing on CI, I'll have a look

tpetricek:
Did you figure this out? @forki replied to my complaints on Twitter saying

> "You should probably use "nuget Fsharp.core redirects:force" and reinstall with Paket."
> and "Don't forget to add it to paket.references"

I have not tried that yet...

ovatsus:
It wasn't a paket issue, I had to change starter pack to make all new arguments optional

Issue #960, opened 2016-07-08
[ Net/Http ] error getting cookies from header response when the specified path is not valid


juicelink:
i'm trying to request an url that contain this set-cookie header in the response
```
set-cookie:_token=en767butjjertvvtbqandb8oa4; expires=Fri, 08-Jul-2016 12:34:17 GMT; Max-Age=3600; path=\
```

the back slash in the path breaks this line in the net /Http.fs file (795):
```
let uri = Uri((if cookie.Secure then "https://" else "http://") + cookie.Domain.TrimStart('.') + cookie.Path)
```

Should it be possible to make a an option, or new function to just get the content response and avoid all other possibilities of failure, for cases like this one where the source is not well formated?

PR #961, opened 2016-07-08
[Http] don't produce cookies from response header if associated uri is not valid

juicelink:
if the set-cookie header in the response is not correct, the request fails.

this change permit to get the response by rejecting cookies with not well formated associated uri

#960


ovatsus:
Thanks

PR #967, opened 2016-07-27
Specify kind on Date header

JonCanning:
> Otherwise it is assumed to be DateTimeKind.Local (local time)

https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest.date(v=vs.110).aspx


ovatsus:
Can you also add a line to the release notes under `2.3.3 - Unrelease`?

PR #968, opened 2016-07-30
Add support for escaped characters in the css selectors.

kontomondo:
Supports `\x` escaping where `x` is any character with the exception of CR, LF, FF and
a hexadecimal digit (per CSS 2.1 spec)

### DIFF
--- a/src/Html/HtmlCssSelectors.fs
+++ b/src/Html/HtmlCssSelectors.fs
@@ -51,6 +51,14 @@ module internal HtmlCssSelectors =
         let getOffset (t:List<char>) = 
             charCount - 1 - t.Length
 
+        let isEscapable (c:char) =
+            (* CSS 2.1: Any character (except a hexadecimal digit, linefeed,
+               carriage return, or form feed) can be escaped with a backslash to
+               remove its special meaning *)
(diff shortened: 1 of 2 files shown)


tauchieta:
Happy to change any code to be more idiomatic to F# and FSharp.Data, I'm new to F#.

ovatsus:
@rflechner, can you review this change?

kontomondo:
Will close this one and re-open with some further changes.

ovatsus:
You can just push a further commit to the same branch and the PR will be updated, you don't need to close and create a new one

rflechner:
Hi,

I ran the tests and reviewed the code.

Everything is working for me but I think we should add a test like:

```fsharp
[<Test>]
let ``special characters can be escaped with class``() =
    let html =
        """
            <!doctype html>
            <html lang="en">
            <body>
            <p id="has:column-and.dot">Some text 1</p>
            <p id="has:column-and.dot" class="toto">Some text 2</p>
            <p id="ignore"/>
            </body>
            </html>
        """
        |> HtmlDocument.Parse
    let selection = html.CssSelect "#has\\:column-and\\.dot.toto"
    selection |> should haveLength 1
    let value = selection |> List.map (fun n -> n.InnerText()) |> Seq.exactlyOne
    value |> should equal "Some text 2"
```

With this test, we ensure that handling escaped chars doesn't break class selectors.



Regards

PR #969, opened 2016-07-30
Add support for escaped characters in the css selectors.

kontomondo:
Supports ` \x` escaping where x is any character with the exception of `CR`, `LF`, `FF` and
a hexadecimal digit (per CSS 2.1 spec).

### DIFF
--- a/src/Html/HtmlCssSelectors.fs
+++ b/src/Html/HtmlCssSelectors.fs
@@ -51,6 +51,14 @@ module internal HtmlCssSelectors =
         let getOffset (t:List<char>) = 
             charCount - 1 - t.Length
 
+        let isCharacterEscapable (c:char) =
+            (* CSS 2.1: Any character (except a hexadecimal digit, linefeed,
+               carriage return, or form feed) can be escaped with a backslash to
+               remove its special meaning *)
+            let isHexadecimalDigit = Char.IsDigit(c) || (Char.ToLower(c) >= 'a' && Char.ToLower(c) <= 'f')
+            (isHexadecimalDigit || c = '\n' || c = '\f' || c = '\r')
+            |> not
+
         let rec readString acc = function
             | c :: t when Char.IsLetterOrDigit(c) || c.Equals('-') || c.Equals('_') 
                 || c.Equals('+') || c.Equals('/')
@@ -62,9 +70,8 @@ module internal HtmlCssSelectors =
                 else
                     inQuotes <- true
                     readString acc t
-            | '\\' :: '\'' :: t when inQuotes ->
-                readString (acc + ('\''.ToString())) t
-
+            | '\\' :: c :: t when isCharacterEscapable c ->
+                readString (acc + (c.ToString())) t
             | c :: t when inQuotes ->
                 readString (acc + (c.ToString())) t
             | c :: t -> acc, c :: t
--- a/tests/FSharp.Data.Tests/HtmlCssSelectors.fs
+++ b/tests/FSharp.Data.Tests/HtmlCssSelectors.fs
@@ -448,5 +448,5 @@ let ``special characters can be escaped``() =
         """ |> HtmlDocument.Parse
     let selection = html.CssSelect ".has-class#has-id-with\\:column-and\\.dot"
     selection |> should haveLength 1
-    let values = selection |> List.map (fun n -> n.InnerText()) |> Seq.exactlyOne
+    let values = selection |> Seq.exactlyOne |> HtmlNode.innerText
     values |> should equal "Matched, has id and class"
--- a/tests/FSharp.Data.Tests/HtmlCssSelectors.fs
+++ b/tests/FSharp.Data.Tests/HtmlCssSelectors.fs
@@ -433,4 +433,19 @@ let ``has attribute Selector``() =
     let values = selection |> List.map (fun n -> n.AttributeValue("data"))
     values |> should equal ["test"]
 
-
+[<Test>]
+let ``special characters can be escaped``() =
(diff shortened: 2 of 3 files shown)


kontomondo:
Happy to change any code to be more idiomatic to F# and FSharp.Data, I'm new to F#.

src/Html/HtmlCssSelectors.fs
```diff
@@ -51,6 +51,14 @@ module internal HtmlCssSelectors =
         let getOffset (t:List<char>) = 
             charCount - 1 - t.Length
 
+        let isCharacterEscapable (c:char) =
+            (* CSS 2.1: Any character (except a hexadecimal digit, linefeed,
+               carriage return, or form feed) can be escaped with a backslash to
+               remove its special meaning *)
+            let isHexadecimalDigit = Char.IsDigit(c) || (Char.ToLower(c) >= 'a' && Char.ToLower(c) <= 'f')
+            (isHexadecimalDigit || c = '\n' || c = '\f' || c = '\r')
```
kontomondo: I see .Equals being used instead below, is there an advantage in doing that? If I do have to change it to .Equals what about the hexadecimal digit comparison above where I need the ascii code ?

kontomondo:
@rflechner Thank you! I've updated the commit to include the changes you suggested for the test. Could you please let me know your thoughts on using `=` comparison instead of .Equals as per above comment of mine? Also since this is my first time working on FSharp.Data, it would be good to make sure that I'm not breaking any naming conventions or any other conventions for the project.

tests/FSharp.Data.Tests/HtmlCssSelectors.fs
```diff
@@ -440,12 +440,13 @@ let ``special characters can be escaped``() =
...
+            <p id="ignore">Ignore</p>
             </body>
             </html>
         """ |> HtmlDocument.Parse
-    let selection = html.CssSelect "#id-has\\:column-and\\.dot"
+    let selection = html.CssSelect ".has-class#has-id-with\\:column-and\\.dot"
     selection |> should haveLength 1
-    let values = selection |> List.map (fun n -> n.InnerText())
-    values |> should equal ["Matched paragraph text"]
+    let values = selection |> List.map (fun n -> n.InnerText()) |> Seq.exactlyOne
```
kontomondo: It would be nice if there was a static HtmlNode.innerText method, this way one could write:
`let values = selection |> Seq.exactlyOne |> HtmlNode.innerText`
If I send a separate pull request for this, would it be considered? Or is it a bad practice to have both static & object method variants?

src/Html/HtmlCssSelectors.fs
ovatsus: = or .Equals is equivalent in this case

src/Html/HtmlCssSelectors.fs
```diff
@@ -62,9 +70,8 @@ module internal HtmlCssSelectors =
                 else
                     inQuotes <- true
                     readString acc t
-            | '\\' :: '\'' :: t when inQuotes ->
```
ovatsus: Don't you need to also keep this rule?

kontomondo: Doesn't the new code supersede it? It would escape the `'` character either in quotes or not in quotes all the same.

ovatsus: but `isCharacterEscapable` will return false for `'`, right?

kontomondo: It will return `true`, there's a `|> not` at the end.

ovatsus: you're right, I missed the not

ovatsus:
Would you mind also adding a line to releasenotes.md about this before I merge?

src/Html/HtmlCssSelectors.fs
kontomondo: @ovatsus It might be a readability issue, is there a better, more idiomatic way to do this? perhaps `not <|` at the beginning or just `not (... )` ?

kontomondo:
@ovatsus Absolutely, what version / release date should I be using ? It seems to be the format used:
``
#### 2.3.x - July 31 2016
* Support for escaping special characters in a CSS selector expression.
```

src/Html/HtmlCssSelectors.fs
smoothdeveloper: `not (isHexadecimalDigit || c = '\n' || c = '\f' || c = '\r')`

ovatsus:
There's already a 2.3.3 item there, just add to the existing one

src/Html/HtmlCssSelectors.fs
kontomondo: @smoothdeveloper  there was already a `|> not` there, or are you saying that `not(..)` style is preferred? I've seen both equally used throughout the F# projects,

smoothdeveloper: Sorry, yes intended to say I'd prefer the not upfront in this specific case.

Yes both can be used liberally, when I use it with pipe in simple expressions though it is most often a sign of my own lazyness (and lack of skills with editor) because I can just append it instead of putting paren back.

I like it used in chain of pipes.

kontomondo:
@ovatsus done

src/Html/HtmlCssSelectors.fs
kontomondo: @smoothdeveloper Thanks for the comment! It already got merged though, but please do change it if you ever touch that file, or I'll change it if I'll have any other PRs.

PR #979, opened 2016-08-15
Fix BuildTests & RunTests on Linux

kontomondo:
BuildTests and RunTests targets are failing due to path resolving issues and some incorrect-case paths.

### DIFF
--- a/tests/FSharp.Data.Tests/Http.fs
+++ b/tests/FSharp.Data.Tests/Http.fs
@@ -16,7 +16,7 @@ open FSharp.Data.HttpRequestHeaders
 let ``Don't throw exceptions on http error`` () = 
     let response = Http.Request("http://api.themoviedb.org/3/search/movie", silentHttpErrors = true)
     response.StatusCode |> should equal 401
-    response.Body |> should equal (Text """{"status_message":"Invalid API key: You must be granted a valid key.","status_code":7}""")
+    response.Body |> should equal (Text """{"status_message":"Invalid API key: You must be granted a valid key.","success":false,"status_code":7}""")
 
 [<Test>]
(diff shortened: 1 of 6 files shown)


kontomondo:
@ovatsus Is it possible to get CI going on Linux as well as Windows?

ovatsus:
We do have CI running on Linux, and it's passing on master

kontomondo:
@ovatsus is there a link I can check ? I can only see Windows & Mac on here and both are case-insensitive by default.

kontomondo:
I see there's http://www.mono-project.com/docs/advanced/iomap/ for mono, so perhaps export MONO_IOMAP=all is set to make it compile/run ? It would be better perhaps to fix the paths in the files so that this wouldn't be required.

tests/FSharp.Data.Tests/Http.fs
```diff
@@ -16,7 +16,7 @@ open FSharp.Data.HttpRequestHeaders
 let ``Don't throw exceptions on http error`` () = 
     let response = Http.Request("http://api.themoviedb.org/3/search/movie", silentHttpErrors = true)
```
kontomondo: If I go to this url with a browser I get:{"status_message":"Invalid API key: You must be granted a valid key.","success":false,"status_code":7} 
Note the addition of success:false in there, which differs from the text in the test. How is this passing in CI?

kontomondo:
@ovatsus Is there anything preventing this from being merged? The build would still fail on Linux without these changes - there's no Linux CI, just Mac & Windows. Thanks!

ovatsus:
We need to setup Linux CI then, otherwise it will be easy to regress again

baronfel:
This should be closed now that #1057 has been merged.

Issue #982, opened 2016-08-17
Http.Request fails with "part of the cookie is invalid" on Google News feed


ptrelford:
When I try to querying Google news:

```
Http.Request("https://news.google.com/news?hl=en&ned=us&ie=UTF8&nolr=1&output=rss&q=FSharp&num=1000")
```

I see the following error:

```
System.Net.CookieException: The 'Name'='' part of the cookie is invalid.
   at System.Net.Cookie.set_Name(String value)
>    at FSharp.Data.CookieHandling.getAllCookiesFromHeader@764-1.Invoke(Int32 i, String cookiePart) in C:\Git\FSharp.Data\src\Net\Http.fs:line 769
   at Microsoft.FSharp.Collections.ArrayModule.IterateIndexed[T](FSharpFunc`2 action, T[] array)
   at FSharp.Data.CookieHandling.getAllCookiesFromHeader@761.GenerateNext(IEnumerable`1& next) in C:\Git\FSharp.Data\src\Net\Http.fs:line 764
   at Microsoft.FSharp.Core.CompilerServices.GeneratedSequenceBase`1.MoveNextImpl()
   at Microsoft.FSharp.Core.CompilerServices.GeneratedSequenceBase`1.System-Collections-IEnumerator-MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at Microsoft.FSharp.Collections.SeqModule.ToArray[T](IEnumerable`1 source)
   at <StartupCode$FSharp-Data>.$Http.InnerRequest@930-5.Invoke(WebResponse _arg2) in C:\Git\FSharp.Data\src\Net\Http.fs:line 930
   at Microsoft.FSharp.Control.AsyncBuilderImpl.args@835-1.Invoke(a a)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.FSharp.Control.AsyncBuilderImpl.commit[a](Result`1 res)
   at Microsoft.FSharp.Control.CancellationTokenOps.RunSynchronously[a](CancellationToken token, FSharpAsync`1 computation, FSharpOption`1 timeout)
   at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken)
```

ajukraine:
I have similar issue. What is the difference between patches?

PR #985, opened 2016-08-18
Fix for Issue #982: "part of the cookie is invalid"

ptrelford:
Fix for function getAllCookiesFromHeader where cookie string is empty no cookie needs to be created.


tpetricek:
As far as I can see, this looks good! 

@ovatsus You're more familiar with the cookie parsing code - can you spot any potential issues here?

Issue #986, opened 2016-08-22
Failing tests in FSharp.Data.Tests\Http.fs


zpbappi:
There are 3 tests in the ~/tests/FSharp.Data.Tests/Http.fs file that are failing. The tests are:
- `Don't throw exceptions on http error`
- `Throw exceptions on http error`
- `Setting timeout in customizeHttpRequest overrides timeout argument`

The reason they are failing has nothing to do with the expected behavior. Rather, the url the tests are calling (http://api.themoviedb.org/3/search/movie) has a new field in the json response named `success`. As the tests are trying to compare the response, the tests are failing. As a result, almost all the pull requests pending fixing these 3 tests by adding the `success` field in the test.

To me, this approach seems like a risky way to assert a behavior. As the json response has nothing to do with the asserted behavior, we can probably re-write the tests without using the json response. I will submit a PR pretty soon with the changes in this 3 tests. However, for these tests, I would feel safer using a mock server that can simulate the behavior we want, without actually relying on the 3rd party web site. Although I am not quite sure how easy that would be in F#. But it would enable us to run tests anytime and anywhere, without even internet connection.

ovatsus:
We already use a mock http server for some tests, see the HttpIntegrationTests.fs

ovatsus:
Already fixed

PR #987, opened 2016-08-22
updated Http tests to not rely on json response from 3rd party

zpbappi:
Fixes the failing tests in Http.fs, addresses issue #986


ovatsus:
Thanks

PR #1000, opened 2016-10-17
Parse time spans

eggapauli:
This PR adds `TimeSpan` as primitive type.  
Basically everywhere `DateTime` is used I added similar logic for `TimeSpan`.
When writing tests I wasn't as ambitious as you (`DateTime` is really used a lot in different tests) and just created tests in `JsonValue` and `JsonProvider`. Some test cases in `InferenceTests.fs` might make sense. what do you say?


dsyme:
@eggapauli Could you bring this up-to-date and clear the CI errors please? thanks

eggapauli:
@dsyme yep, thanks for notifying. Will do it soon.

eggapauli:
see #1023

Issue #989, opened 2016-08-24
Dr. Who in trouble


zpbappi:
Interestingly, the test `viewersByDoctor` in `docs/content/library/HtmlProvider.fsx` (line 130 at the time of writing) is failing in AppVeyor and Travis-CI. However, it passes nicely in my machine. Even the following script runs perfectly:

``` fsharp
open FSharp.Data
open System

let [<Literal>] DrWho = 
      "http://en.wikipedia.org/wiki/List_of_Doctor_Who_serials"
let test () =
    let doctorWho = new HtmlProvider<DrWho>()

    doctorWho.Tables.``Series overview``.Rows
    |> Seq.map (fun r -> r.``Viewers (millions) - Finale``)
    |> Seq.iter Console.WriteLine

test()
```

I can only find the following detail from [travis-ci logs](https://travis-ci.org/fsharp/FSharp.Data/builds/154637656):

``` text
1) Failed : FSharp.Data.DesignTime.Tests.DocumentationTests.Documentation generated correctly ("library/HtmlProvider.fsx")
Found errors when processing file 'library/HtmlProvider.fsx':
((136, 19), (136, 50)) The field, constructor or member 'Viewers (millions) - Finale' is not defined (library/HtmlProvider.fsx)
at FSharp.Data.DesignTime.Tests.DocumentationTests.Documentation generated correctly  (System.String file) in <filename unknown>:line 0
```

Anyone with any idea about why this is happening?

taylorwood:
There have definitely been times before when Wikipedia's Dr. Who page changes and this test needed to be updated, but it's odd that it passes on your machine but not another.

zpbappi:
One possible reason I can think of is wiki in my region (Sydney) is not updated yet, whereas it is updated in other regions. But it seems unlikely as the test has been failing for last few days in CI servers, but running nicely in my machine. I don't think wiki have a different version for Australia. If page update is the case, can we slowly migrate all the tests to use mock server to remove the dependecies on external systems? Possibly starting with the failing ones.

tpetricek:
I think we also do some caching of schema in type providers (possibly!)

zpbappi:
_**tl;dr** EXPLAIN... EXPLAIN..._

[...]

PR #1003, opened 2016-10-30
Fix for issue #989: Dr. Who in Trouble

jschiefer:
Issue #989 is about the fact that the `Documentation generated correctly ("library/HtmlProvider.fsx")` test fails, because the `Viewers (millions) - Finale` column is missing. It is missing because the column count of the `Series overview` table  on https://en.wikipedia.org/wiki/List_of_Doctor_Who_serials is off by one. If you look at this table closely, you will notice that all the cells in the "Episodes" column have a colspan of 2, for no obvious reason. This causes the column count too be low by 1.

So the fix is to account for the colspans when determining the number of columns in a table.

### DIFF
--- a/src/Html/HtmlRuntime.fs
+++ b/src/Html/HtmlRuntime.fs
@@ -149,6 +149,10 @@ module HtmlRuntime =
             i
 
     let private parseTable inferenceParameters includeLayoutTables makeUnique index (table:HtmlNode, parents:HtmlNode list) = 
+        let rowSpan cell = 
+            max 1 (defaultArg (TextConversions.AsInteger CultureInfo.InvariantCulture cell?rowspan) 0) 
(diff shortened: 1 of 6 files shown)


zpbappi:
anyone: travis-ci log says it can't find mono! But, it was installed successfully in the previous step. Path issue, may be?
> ./build.sh: line 21: mono: command not found

@jschiefer you probably forgot to change signature files. Look at the 3 failed appveryor tests.

jschiefer:
I didn't change those, because they weren't related to the problem I was fixing. However, I guess a happier AppVeyor is better... Just added the fix.

ovatsus:
Thanks

dsyme:
I just love the title to this issue :)

PR #1004, opened 2016-11-01
xsd support

giacomociti:
The XML Provider is augmented with XSD support as in [FSharp.Data.Xsd](https://fsprojects.github.io/FSharp.Data.Xsd/).
An new parameter `Schema` is added. When this parameter is used (instead of `Sample`), an `InferedType` is derived from the specified XSD. 

In order to leverage the existing implementation, the approach (as made evident in the unit tests) is to derive a type essentially equivalent to the one that would be inferred from a significant set of valid samples. The focus is on elements and not on complex types: provided types correspond to element shapes, not to complex types in the schema.

I may also provide a bit of documentation (English only, no Japanese :)


kerams:
I'd love to see this merged. @ovatsus or @tpetricek, could you please take a look? :)

Vandroiy:
This should get more attention! I've tried this shortly after this PR was made and found it a big improvement. I've migrated any XML code I worked on since to the XSD provider, which led to shorter and cleaner F# code, removed the need for sample files, improved the quality of provided types, and made me notice errors in my XSDs. It would be neat if this got merged, so that issues dependent on FSharp.Data, like proper single-precision floating-point support, can get resolved.

ovatsus:
Sorry, it's on my todo list (together with a bunch of other PRs), but just haven't had the time lately to review and merge

giacomociti:
thanks for the feedback. I'm available for clarification of details but also about the overall design

ovatsus:
#57

baronfel:
@giacomociti do you have time to rebase this? If not I can rebase it and submit a new PR.  This is great work and we should get it in :)

giacomociti:
@baronfel  Please, rebase and submit a new PR (I'm having issues with my PC at the moment)
Thanks

PR #1019, opened 2017-01-10
Improved performance when parsing CDATA sections

mrange:
My use-case involves parsing a HTML document with "large" CDATA sections (80KiB).

The existing implementation seemingly hanged but in truth was just progressing very slowly.

The problem seemed to be that for every CDATA character this was performed:

```fsharp
if ((!state.Content).ToString().EndsWith("]]>"))
````

ToString is implemented like this:

```fsharp
override x.ToString() = String(x.Contents |> List.rev |> List.toArray)
```

So for each CDATA character we reverse the current content, makes it into an array and then creates a string.

Instead I changed cData to accept a parameter which controls whether an end tag is being processed.

In addition, I added a unit test for the performance for parsing large CDATA sections. Also, it seems that the CDATA sub parser wasn't triggered for the current unit test (script tag seems to have special handling) so I modified the existing test.

I ran the tests with old and new parser and the only difference is that when I run the large CDATA test it doesn't finish in time.

For my use-case parsing large CDATA section is very important.

Regards,

Mårten

### DIFF
--- a/src/Html/HtmlParser.fs
+++ b/src/Html/HtmlParser.fs
@@ -549,21 +549,23 @@ module internal HtmlParser =
             | current -> 
                 match new String(Array.append current (state.Pop(5))) with
                 | "DOCTYPE" -> docType state
-                | "[CDATA[" -> state.Cons("<![CDATA[".ToCharArray()); cData state
+                | "[CDATA[" -> state.Cons("<![CDATA[".ToCharArray()); cData 0 state
                 | _ -> bogusComment state
-        and cData (state:HtmlState) = 
-            if ((!state.Content).ToString().EndsWith("]]>"))
-            then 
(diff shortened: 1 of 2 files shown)


mrange:
`Travis CI` complains about: 

> ./build.sh: line 21: mono: command not found

ovatsus:
@colinbull can you review as you're more familiar with this part of the code?

colinbull:
O.K, tested this this looks good to me. 

@mrange  Accidentally sent you a PR.. Basically because I was being stupid.

@ovatsus I'm going to go ahead and merge this and fix the travis build afterwards. Unless you have any objections.

ovatsus:
Go ahead and merge

mrange:
Out of interest; what is stopping the merge?

PR #1020, opened 2017-01-16
Try to resurrect Travis CI

dsyme:
Travis seems toast, this tries to resurrect

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,10 @@ language: csharp
 matrix:
   include:
     - os: osx
-      mono: 4.2.0
-    - os: osx
       mono: 4.4.2
     - os: osx
       mono: latest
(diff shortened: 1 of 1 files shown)


dsyme:
(The matrix does a 6x build and test, which may be too much by a long way, we can trim it down)

dsyme:
Modeled on https://github.com/Microsoft/visualfsharp/blob/master/.travis.yml, though seems to give an incorrect matrix right now

dsyme:
Adjusted the matrix a bit to only use OSX.  Compilation on Linux fails because of a lack of installed FSharp.Core PCL libraries.  This should be fixed by having a dependency on the FSharp.Core nuget package (which contains the PCL assemblies) rather than assuming they are installed as part of Mono.

PR #1030, opened 2017-02-10
Fix cssselector to work outside the body element

vladkosarev:
Added  .Html and tryGetHtml functionality so that css selectors can be done from html tag level instead of body level


ovatsus:
Thanks

PR #1042, opened 2017-04-11
Update FAKE version

forki:
see https://github.com/fsharp/FAKE/issues/1515


BlythMeister:
I think the targets at the bottom also need updating.
It could still try running tests before build on the  "All" target.

ovatsus:
I think I prefer the old behaviour and will wait for the FAKE fix

forki:
keeping this for testing

forki:
looks like FAKE is back to normal

ovatsus:
Great!

PR #1043, opened 2017-04-11
Use Paket in Magic Mode

forki:


ovatsus:
What's magic mode?

forki:
https://fsprojects.github.io/Paket/bootstrapper.html#Magic-mode

ovatsus:
ah, got it

PR #1044, opened 2017-04-11
Revert "Update FAKE version"

ovatsus:
Reverts fsharp/FSharp.Data#1042

PR #1045, opened 2017-04-11
Force FAKE to use MSBuild 14

ovatsus:


build.sh
```diff
@@ -18,5 +19,7 @@ else
   if [ $exit_code -ne 0 ]; then
   	exit $exit_code
   fi
+
+  export VisualStudioVersion=14.0
```
ForNeVeR: I m sorry to bother you, but according to [@forki comment](https://github.com/fsharp/FSharp.Data/pull/1044#issuecomment-293246058), this environment variable is checked only on Windows. There's no sense in setting it on Linux build.

PR #1049, opened 2017-04-23
docs: replace FSharp.Net with FSharp.Data

ForNeVeR:
F# Russian Community users have noticed that the comment is inconsistent with the corresponding code. It seems that after moving the functionality to `FSharp.Data` namespace somebody forget to change the documentation accordingly.

_Note_: I don't know Japanese, but I can clearly see the same typo in the Japanese translation, so I've decided to fix that, too.


ovatsus:
Thanks

PR #1051, opened 2017-05-10
[WIP] test on Mono 5

sergey-tihon:
Test that FSharp.Data work on Mono 5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ matrix:
       mono: 4.4.2
     - os: osx
       mono: latest
+    - os: osx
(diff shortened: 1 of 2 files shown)


sergey-tihon:
Close in favor of #1056

baronfel:
This is going to be a bit of work to make work on mono 5, because the reference assemblies are just not there at all on OSX. on linux the situation seems even more dire: the `referenceassemblies-pcl` package doesn't seem to drop the assemblies in the `Reference Assemblies` folder off of the root mono dir :(

Issue #1054, opened 2017-05-12
Please update ProvidedTypes.fs for Mono 5.0 


dsyme:
Please update to use the latest version of ProvidedTypes.fs, as described in  https://github.com/fsprojects/FSharp.TypeProviders.StarterPack/issues/121

baronfel:
This should be in fixed in PR #1056 as well as part of a general paket update

baronfel:
This can be closed now, I believe.

PR #1057, opened 2017-06-23
Build and test fixes

baronfel:
This PR pull out parts of the multipart form data PR that fix the following issues:

Broken Tests on travis (typically due to casing or yahoo's stock API being deprecated)
Adding linux support to the build (mostly requires package installs and path locations in the test helper modules)
Package updates (including the mono 5 type provider fix)
Paket-referencing FSharp.Core (which helps solve at least two of the above issues :))


baronfel:
Handles #1054 and #1051, at a minimum.

baronfel:
anyway, @ovatsus linux support is super close. There's some discrepancy between the search.atom.xml file on the file system and in the tests, and the linux runners for the current HEAD of this branch are having errors starting up, but the OSX builds are good 👍

dsyme:
@baronfel Fantastic work! :)

.travis.yml
```diff
@@ -1,30 +1,38 @@
 language: csharp
 
+sudo: required
+dist: trusty
+
+mono:
+  - 4.4.2
```
dsyme: Can you add "latest" and even "nightly" here too?  Perhaps drop one or two of the others

baronfel:
holy smokes all the casing issues are fixed up now! we're successfully building and testing on linux!

baronfel:
as far as I am concerned this PR is complete.

ovatsus:
Thanks, I'll try to review and merge later today

ovatsus:
Since we're now referencing FSharp.Core from NuGet, shouldn't we also update the nuspec so consumers of FSharp.Data also consume FSharp.Core from NuGet, otherwise they can have mismatches, no?

baronfel:
That's a good question.  In the previous version, we built and deployed against FSharp.Core version 4.3.0.0, and any users on further-ahead versions had to do binding redirects.  I think if anything, I should update the nuspec to show that we have an FSharp.Core dependency and let clients at least have that truth, instead of not directly exposing our dependency on FSharp.Core.

baronfel:
And now that I read what you said on my laptop and not my phone, I realize that's exactly what you said.

baronfel:
can you sanity check that nuspec change for me? I haven't found a way to extract the nuget package on OSX yet to verify the contents

ovatsus:
CI was green on the PR, but after merging the CI is failing on Mono 5.01 on Mac

baronfel:
I'll take a look

baronfel:
Seems to be the same intermittent DNS failure that plagued a few of the
Linux builds on the PR. Can you retrigger the build?

ovatsus:
Yep, that was it, after retrying it passed

PR #1061, opened 2017-07-21
Fix #1022: HtmlProvider Self closing table tags

### LINKED ISSUE #1022: HtmlProvider Self closing table tags
HtmlProvider doesn't provide tables if they contain self closing tags which are not closed.

[According to w3.org HTML 4.01](https://www.w3.org/TR/html401/struct/tables.html#edef-TD) and [HTML 5](https://www.w3.org/TR/html5/tabular-data.html#the-td-element) TD, TH TR tags needn't be closed. Sadly many apps take advantage of this.
 ## Expected
Parsing self closing tags should behave the same regardless of whether the tags are closed.
 ## Sample error throwing code
```fsharp
// Missing </th>, </tr> and </td>
let [<Literal>] Example = """
<!DOCTYPE html>
<html>
  <body>
    <table>
      <tr>
        <th>Number
        <th>Text
      <tr>
        <td>4
        <td>Hello
    </table>
  </body>
</html>"""

type NotWorking = HtmlProvider<Example>
(NotWorking.GetSample()).Tables.Table1.Rows
```
Fails with
```
Error #39 - The field, constructor or member 'Tables' is not defined at line 27 col 25
```
 ## Sample working alternative
```html
<!DOCTYPE html>
<html>
  <body>
    <table>
      <tr>
        <th>Number</th>
        <th>Text</th>
      </tr>
      <tr>
        <td>4</td>
        <td>Hello</td>
      </tr>
    </table>
  </body>
</html>
```

The html files are equal, but the second works while the first one does not.

thinkbeforecoding:
This PR fixes #1022 and also implicitly closed li list items, like the following
```
<ul>
   <li>1
   <li>2
   <li>3
</ul>
```

### DIFF
--- a/docs/content/library/HtmlProvider.fsx
+++ b/docs/content/library/HtmlProvider.fsx
@@ -129,7 +129,7 @@ let doctorWho = new HtmlProvider<DrWho>()
 // Get the average number of viewers for each doctor's series run
 let viewersByDoctor = 
   doctorWho.Tables.``Series overview``.Rows 
-  |> Seq.groupBy (fun season -> season.``Doctor(s)``)
+  |> Seq.groupBy (fun season -> season.Doctor)
   |> Seq.map (fun (doctor, seasons) -> 
       let averaged = 
         seasons 
--- a/src/Html/HtmlParser.fs
+++ b/src/Html/HtmlParser.fs
@@ -704,6 +704,43 @@ module internal HtmlParser =
             | "area" | "base" | "br" | "col" | "embed"| "hr" | "img" | "input" | "keygen" | "link" | "menuitem" | "meta" | "param" 
             | "source" | "track" | "wbr" -> true
             | _ -> false
+
+        let isImplicitlyClosedByStartTag expectedTagEnd startTag  =
+            match expectedTagEnd, startTag with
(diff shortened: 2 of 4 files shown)


thinkbeforecoding:
The build is failing on the EBay car Html page signature.
It finds an additional ```Additional site navigation``` Table that was not found before because it was not well formed...
How should we handle this change ?

dsyme:
@thinkbeforecoding Fix the test - it's sounds like you've fixed a bug

thinkbeforecoding:
Ok. I changed the expected result since the new one should be better

thinkbeforecoding:
Previously, when starting new td, or new li without closing the new one, it was nesting them.. and adding a lot of closing tags at the end.
This prevented the HtmlType provider from finding the structure. Now the structure stays as expected and the type provider finds the tables or lists 👍

thinkbeforecoding:
Haha, the build failed because there was an edit on the Dr Who wikipedia page !
The tables was named 'Doctor(s)' and it's now 'Doctor'

thinkbeforecoding:
@dsyme Cool. Everything is compiling and passing tests 👍

ovatsus:
👍

thinkbeforecoding:
😃

PR #1067, opened 2017-09-01
Add `JsonValue.TryParse`

mjgpy3:
Perhaps the dumbest PR so I understand if this is rejected but, that being said, here it is!

My reasons for submitting it are
 - Wrapping in option helps with an functional style, rather than having exceptions and, selfishly
 - I write this same wrapper in every project that uses FSharp.Data.

Let me know your thoughts!

### DIFF
--- a/tests/FSharp.Data.Tests/JsonValue.fs
+++ b/tests/FSharp.Data.Tests/JsonValue.fs
@@ -201,6 +201,16 @@ let ``Can parse nested array``() =
     j.[1].[1] |> should equal (JsonValue.String "Clyde")
 
 [<Test>]
+let ``TryParse is None when a bad JSON value is given``() =
+    let j = JsonValue.TryParse "}{"
+    j |> should equal None
+
+[<Test>]
+let ``TryParse is Some of the correct JSON value when a good structure is given``() =
+    let j = JsonValue.TryParse """{ "foo": "bar" }"""
+    j |> should equal (Some (JsonValue.Record [| "foo", JsonValue.String "bar" |]))
+
+[<Test>]
 let ``Can serialize empty document``() =
     (JsonValue.Record [| |]).ToString(JsonSaveOptions.DisableFormatting)
     |> should equal "{}"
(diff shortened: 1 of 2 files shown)


realvictorprm:
Actually not that worse I think, if you have some use-case for this it should be fine because it doesn't breaks anything nor it infers wrong usage.

tests/FSharp.Data.Tests/JsonValue.fs
```diff
@@ -416,4 +426,4 @@ let ``Encoding of markup is not overzealous`` () =
   let w = new IO.StringWriter()
   JsonValue.JsonStringEncodeTo w input
   let expected = input // Should not escape </>
-  (w.GetStringBuilder().ToString()) |> should equal expected
\ No newline at end of file
+  (w.GetStringBuilder().ToString()) |> should equal expected
```
realvictorprm: Hm, could you please remove the newline afterwards? It's unnecessary here 😕

realvictorprm:
@tpetricek sorry Sir for talking to you that much today but I struggled over this PR. Just wanted to know what you think, is this ready to merge?

realvictorprm:
Sure, I'ld try my best @tpetricek.
I noticed that writing parsing stuff with F# is much fun and something I'm definitely continuing :smile:

tests/FSharp.Data.Tests/JsonValue.fs
mjgpy3: @realvictorprm, this is done.

mjgpy3:
@realvictorprm, @tpetricek I've squashed the commits but I can't seem to build FSharp.Data locally (even just from master). Any ideas as to why the appveyour is failing?

realvictorprm:
@mjgpy3 the build worked for me yesterday, however I noticed that the some HttpRequest tests on the CI fail. However this is very unrelated to your addition.

mjgpy3:
@realvictorprm I see, thank you.

ovatsus:
I think this is fine for now, but long term we should change the parser so we don't need to throw and catch. And maybe could output in the failure case the error message stating where the parsing failed (or just the char numbers directly)?

realvictorprm:
Hm, I could take care of this after I added the schema support

PR #1068, opened 2017-09-14
Update Year

realvictorprm:
I think it's fine to update the year, right?


tpetricek:
Sounds good :) maybe we should have gone for 2018 already!

PR #1070, opened 2017-09-15
Fix unit tests on datetime conversion in another timezone than GMT

devcrafting:


devcrafting:
A unit test was failing around `case "2016-11-21T13:29:05+03:00" <| System.DateTime(2016,11,21,10,29,05, System.DateTimeKind.Utc)` in TextConversion file, when run in another timezone than GMT. 

The actual value was `System.DateTime(2016,11,21,11,29,05, System.DateTimeKind.Local)` for me (CEST timezone). It is due to the use of DateTimeStyles.RoundtripKind, which keep the DateTimeKind from the parsed string. In GMT timezone, local time and GMT are the same value.

We saw that with @dmitrybatalov during F# mentorship program. He experienced the same error from Russia.

We then switched to DateTimeStyles.AdjustToUniversal. Tested in GMT and CEST timezones (but can't write an automated test for this...).

devcrafting:
Also add git ignore on log files...

realvictorprm:
I'll look into this later. If I forget it, feel free to remind me ;)

DmitryBatalov:
Hey @realvictorprm , @devcrafting and I during F# mentorship program were experimenting with that PR. I'm pinging you :)

realvictorprm:
well okay I would like to see a test but it isn't possible. 
Anyways this looks fine for me. @tpetricek you can merge if you want.

realvictorprm:
Many thanks for the reminder @DmitryBatalov, I've looked into it.

ovatsus:
I remember this being changed to RoundtripKind before to fix some issue, so let's dig that up first and make sure this won't regress anything

realvictorprm:
Hm maybe we can add a parameter to the function so the user can select?

ovatsus:
I'd rather try to understand exactly what option is needed and when it is needed before just asking for an extra decision on the users, since this code is indirectly used by all the type providers. DateTime's are tricky, so I want to investigate thoroughly before changing this again

devcrafting:
We (with @DmitryBatalov) tried to figure out what could be the rationale to use RoundtripKind...unfortunately there is no test breaking when we change it to AdjustToUniversal...

For test, we need to change the OS timezone, so we though to do it in [appveyor](http://help.appveyor.com/discussions/problems/2391-setting-build-worker-time-zone) or [travis](https://stackoverflow.com/questions/23371542/how-do-i-configure-travis-ci-to-use-the-correct-time-zone-for-a-rails-app/23374438).

I have no idea of this bug's impact for users not in GMT timezine. Does anyone have an idea?

ovatsus:
So, this was the commit where it changed from AdjustToUniversal to RoundTripKind - https://github.com/fsharp/FSharp.Data/pull/58/files#diff-32d1e283853a0052822ed289be7315d7 - but that doesn't help understand why it was changed :/ I'll change my timezone locally and do some more testing

ovatsus:
I looked into it a bit more, and I believe the code is correct, just the test needs to be adjusted. AsDateTime is returning the date in local time, and has done the correct conversion. When my PC is on UTC+1, it returns  11:29m correctly

ovatsus:
Fixed here: https://github.com/fsharp/FSharp.Data/commit/5d3deac2728187d2eef8a671e3b5191603a64a5d

PR #1071, opened 2017-09-15
The expected file for World Bank was updated

DmitryBatalov:
The Wold Bank data was updated, so expected file should be updated as well.


ovatsus:
Thanks

Issue #1083, opened 2017-10-09
German date parsing in CSV provider


forki:
Following code:

    let [<Literal>]MesspreisregelungenFile = __SOURCE_DIRECTORY__ + "/data/preisregelungen.csv"
    type Messpreisregelungen = CsvProvider<MesspreisregelungenFile, ";", InferRows = 0, Culture = "de-DE">
      

and following csv:

  
  
  
  Preisregelung_ID;Messgebiet_Nr;gueltig_seit;gueltig_bis;ID;Status_ID;Internet;Bemerkung;Erfasser;Ersterfassung;letzte_Pruefung;letzte_Aenderung
      1;184370001;01.01.2006;30.09.2007;3;2300;;;1;27.09.2006;11.07.2008;11.07.2008
      2;214230001;01.02.2006;20.03.2007;2;2000;;;1;27.09.2006;28.11.2007;10.04.2007

gueltig_seit and gueltig_bis are dates, but CsvProvider detects decimal!?

Decimal is pretty aweful since  01.01.2006 is 1012006M and therefor smaller than 31.12.2005 which parses as 31122005M

forki:
I took a look at textConversion:FS: we have the following code:

      let dateTimeStyles = DateTimeStyles.AllowWhiteSpaces ||| DateTimeStyles.RoundtripKind
      match DateTime.TryParse(text, cultureInfo, dateTimeStyles) with
      | true, d ->
          if d.Kind = DateTimeKind.Unspecified then 
            new DateTime(d.Ticks, DateTimeKind.Local) |> Some
          else 
            Some d
      | _ -> None

I verified that in the REPL. 01.01.2006 gives `Some 01.01.2006 00:00`

baronfel:
Could the problem be in [`StructuralInference.fs`](https://github.com/fsharp/FSharp.Data/blob/master/src/CommonRuntime/StructuralInference.fs#L281)?  The active patterns here, if the ordering is to be believed, would parse as decimals the dates you've provided.

baronfel:
Eh, nevermind that. I see that all the active patterns take the culture as a parameter.

baronfel:
aha, that _is_ it!

```
ci;;
val it : CultureInfo =
  de-DE {Calendar = System.Globalization.GregorianCalendar;
         CompareInfo = CompareInfo - de-DE;
         CultureTypes = AllCultures;
         DateTimeFormat = System.Globalization.DateTimeFormatInfo;
[...]

PR #1084, opened 2017-10-09
Create a failing test for #1083

forki:
#1083


ovatsus:
Thanks

forki:
thanks for merging: just tested it works on my real data

Issue #1081, opened 2017-10-06
Add TryFindColumn functionality to CsvFile


ovatsus:
The idea is to be able to do `csv.TryFindColumn "columnName"`

Today that can be done with `csv.Headers |> Option.bind (Array.tryFindIndex ((=) "columnName"))`

PR #1092, opened 2017-10-18
Add CsvFile.TryFindColumn for #1081

DanielOliver:
I pretty much used the code given in the issue #1081 and added two tests (hopefully in the correct spot).


ovatsus:
Actually, I just realised there's an existing GetColumnIndex but it's internal. Maybe we would make it public and add this one as TryGetColumnIndex? And I think we should add them only on the non TP CsvFile version

tests/FSharp.Data.Tests/CsvProvider.fs
```diff
@@ -563,3 +563,15 @@ let ``Having null in a cell should not fail saving to string (issue#978)`` () =
     data
     |> Stringify 
     |> ignore
+    
+[<Test>]
+let ``CsvFile.TryFindColumn returns only the first matching column``() =
```
ovatsus: Don't forget to update the test name

DanielOliver: Yep. serves me right for rushing this between things today.

ovatsus:
Thanks

PR #1093, opened 2017-10-19
Additional parsing for scripts

Regent:
Whilst investigating original issue I've found a few more cases when parser would mistake special characters for end tag sequence.

### DIFF
--- a/src/Html/HtmlParser.fs
+++ b/src/Html/HtmlParser.fs
@@ -462,6 +462,7 @@ module internal HtmlParser =
             match state.Peek() with
             | TextParser.EndOfFile _ -> data state
             | '/' -> state.Cons(); script state
+            | '\\' -> state.Cons(); state.Cons(); scriptRegex state
             | _ -> state.Cons(); scriptRegex state
         and scriptLessThanSign state =
             match state.Peek() with
--- a/src/Html/HtmlParser.fs
+++ b/src/Html/HtmlParser.fs
@@ -423,8 +423,40 @@ module internal HtmlParser =
         and script state =
             match state.Peek() with
             | TextParser.EndOfFile _ -> data state
+            | ''' -> state.Cons(); scriptSingleQuoteString state
+            | '"' -> state.Cons(); scriptDoubleQuoteString state
+            | '/' -> state.Cons(); scriptSlash state
             | '<' -> state.Pop(); scriptLessThanSign state
             | _ -> state.Cons(); script state
+        and scriptSingleQuoteString state =
+            match state.Peek() with
+            | ''' -> state.Cons(); script state
+            | _ -> state.Cons(); scriptSingleQuoteString state
+        and scriptDoubleQuoteString state =
+            match state.Peek() with
+            | '"' -> state.Cons(); script state
+            | _ -> state.Cons(); scriptDoubleQuoteString state
+        and scriptSlash state =
+            match state.Peek() with
+            | '/' -> state.Cons(); scriptSinleLineComment state
+            | '*' -> state.Cons(); scriptMultiLineComment state
+            | _ -> state.Cons(); scriptRegex state
+        and scriptMultiLineComment state =
+            match state.Peek() with
+            | '*' -> state.Cons(); scriptMultiLineCommentStar state
+            | _ -> state.Cons(); scriptMultiLineComment state
+        and scriptMultiLineCommentStar state =
+            match state.Peek() with
+            | '/' -> state.Cons(); script state
+            | _ -> scriptMultiLineComment state
+        and scriptSinleLineComment state =
+            match state.Peek() with
+            | '\n' -> state.Cons(); script state
+            | _ -> state.Cons(); scriptSinleLineComment state
+        and scriptRegex state =
+            match state.Peek() with
+            | '/' -> state.Cons(); script state
+            | _ -> state.Cons(); scriptRegex state
         and scriptLessThanSign state =
             match state.Peek() with
             | '/' -> state.Pop(); scriptEndTagOpen state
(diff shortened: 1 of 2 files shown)


src/Html/HtmlParser.fs
```diff
@@ -423,8 +423,40 @@ module internal HtmlParser =
...
+            match state.Peek() with
+            | ''' -> state.Cons(); script state
+            | _ -> state.Cons(); scriptSingleQuoteString state
+        and scriptDoubleQuoteString state =
+            match state.Peek() with
+            | '"' -> state.Cons(); script state
+            | _ -> state.Cons(); scriptDoubleQuoteString state
+        and scriptSlash state =
+            match state.Peek() with
+            | '/' -> state.Cons(); scriptSinleLineComment state
```
ovatsus: Typo: sinle instead of single

tests/FSharp.Data.Tests/HtmlParser.fs
```diff
@@ -96,6 +96,18 @@ let ``Can handle attributes with no value``() =
         ]
     node.Attributes() |> should equal expected
 
+[<TestCase("var r =\"</script>\"")>]
+[<TestCase("var r ='</script>'")>]
+[<TestCase("var r =/</g")>]
```
ovatsus: Can you add a test case with "var r =/\/</g"

tests/FSharp.Data.Tests/HtmlParser.fs
```diff
@@ -96,6 +96,18 @@ let ``Can handle attributes with no value``() =
...
+[<TestCase("//</script>\n")>]
```
ovatsus: Can you add the opposite case, not having the \n at the end to make sure it fails

tests/FSharp.Data.Tests/HtmlParser.fs
```diff
@@ -96,6 +96,18 @@ let ``Can handle attributes with no value``() =
...
+[<TestCase("/*</script>*/")>]
+[<TestCase("/*</script>**/")>]
```
ovatsus: Can you add test cases that use `/* */` and span multiple lines?

ovatsus:
@Regent , is this finished now or are you still looking at more corner cases? @colinbull , can you have a quick look as well?

Regent:
@ovatsus, in coming days I also want to cover the case of parsing `<script>0;<body></body>` correctly - at the moment it would parse as simply `<script />` - but ideally I would like to send a separate PR for that.

ovatsus:
Sorry, I thought I had already merged this before

ovatsus:
Released in 2.4.3

glchapman:
Doing some scraping, I found a case where it appears this change (or at least changes since 2.4.2) do not find the end tag of a script. The url in question is :
https://www.marketwatch.com/investing/stock/msft/analystestimates
(where any stock ticker can replace "msft").

I'm attaching captures from the VS Debugger's text visualization of the script node's contents (Raw View).  As you can see, in 2.4.2, the script node is correctly terminated, whereas in 2.4.6, it is not terminated until the closing tag of the ensuing script.

[Script_2.4.2.txt](https://github.com/fsharp/FSharp.Data/files/1936017/Script_2.4.2.txt)
[Script_2.4.6.txt](https://github.com/fsharp/FSharp.Data/files/1936026/Script_2.4.6.txt)

glchapman:
I believe I've narrowed down my problem to the handling of regexes in scripts. As you can see in the attached example, the regex contains a '/' in a bracketed character class. The tokenizer mistakes this for the end delimiter of the regex, and so fails to tokenize correctly.
[test_html.txt](https://github.com/fsharp/FSharp.Data/files/1939269/test_html.txt)
Here's fsi output for 2.4.6 with the above test.html:
```
> let doc = HtmlDocument.Load("test.html");;
val doc : HtmlDocument =
  <html>
  <head />
  <body>
    <script type="text/javascript">
  function(selector){
    return selector.replace(/([/.])/g, '\\$1');
  };
</script>
<div>captured by script</div>
<script></script>
  </body>
</html>

> let script = HtmlDocument.descendantsNamed false ["script"] doc |> Seq.head;;
val script : HtmlNode =
  <script type="text/javascript">
  function(selector){
    return selector.replace(/([/.])/g, '\\$1');
  };
</script>
<div>captured by script</div>
<script></script>
```

ovatsus:
@glchapman , can you send a PR with a failing test case please? And if you could also fix it that would be great, but if you send the failing tests that's already very helpful

Regent:
I haven't realised that you don't need to escape forward slash when it is inside a group. I can't find a definitive specification on the matter - hoped to get it from www.ecma-international.org but they might have forgotten to pay for the hosting...

glchapman:
I'll try to send a PR with a failing case in a bit, but FWIW my reading of the current HTML standard implies you don't actually have to parse the script text (to handle regexes, etc). Instead, raw text nodes (the content of scripts and styles) can have any text at all up to the terminating tag:
https://www.w3.org/TR/2017/REC-html52-20171214/syntax.html#restrictions-on-the-contents-of-raw-text-and-escapable-raw-text-elements
See also the description of parsing the RAWTEXT state:
https://www.w3.org/TR/2017/REC-html52-20171214/syntax.html#rawtext-state

I don't know if this differs from earlier HTML variants.

colinbull:
I'm just looking at this now. @glchapman Can you let me push to your PR with the failing test?

glchapman:
@colinbull  -- I've added you as a collaborator to my fork with the PR. Is this what you need? (I'm sorry, I'm not a big github user).

colinbull:
Yep thanks. Ignore my comments before I have caught up on whats going on now.

Regent:
Original motivation was to handle some corner-cases, e.g. so that </ within regex doesn't get parsed as start of an end tag.

But that's a good point - do we need to parse scripts at all. If I read it correctly, HTML spec simply looks for a closing script tag.
One good thing about current implementation is that it's resilient to the lack of closing tag.

colinbull:
Ok, I have the unit tests passing now but have broken the signature tests (ebay_cars.htm). I haven't got anymore time tonight to look but hoping to get back to this, tomorrow or wednesday.

PR #1095, opened 2017-10-24
Updated HttpCssSelector documentation

Regent:
Tried using `https://www.packtpub.com/all?search=f%23` but have rather inconvenient DOM structure in my opinion.


ovatsus:
Thanks

PR #1103, opened 2017-11-26
Fix incorrect cookie value when it contains '=' sign

devcrafting:
Http.fs cooking handling was calling split('='), removing all '=' from the string whereas the goal was just to split name and value. This bug was introduced in PR #945.

I added a test and reused code before #945 searching for firstEqual with indexOf. I kept a guard if it was not found (since it was the logic implemented with split('=')), but not sure to understand what is the intent (cookie with only a name ?)...

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1465,11 +1465,12 @@ module internal CookieHandling =
             |> Array.iteri (fun i cookiePart ->
                 let cookiePart = cookiePart.Trim()
                 if i = 0 then
-                    let kvp = cookiePart.Split '='
-                    if kvp.Length > 0 then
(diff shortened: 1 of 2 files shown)


ovatsus:
Thank you

ovatsus:
Released in 2.4.3

devcrafting:
Any news about accepting/rejecting this PR ? Thanks

ovatsus:
It was already merged and released...

On 5 Dec 2017 19:51, Clément Bouillier <notifications@github.com> wrote:

Any news about accepting/rejecting this PR ? Thanks

devcrafting:
Great and thanks! Sorry it was just a page refresh problem (I thought GitHub was auto-refreshing...)

Issue #1082, opened 2017-10-08
[Http.fs bug] error with uploading request body


saneman1:
I do PUT http request with some  body, size about 1MB.
The server immediately responds with error status code (409 Conflict), I see it in the traces. After upload of the request body is complete, an error occurs:

WebException: The underlying connection was closed: An unexpected error occurred on a receive.
  IOException: Unable to read data from the transport connection

exception raised at function [getResponseFromBeginEnd](https://github.com/fsharp/FSharp.Data/blob/master/src/Net/Http.fs#L1311)

When size of payload request body less 1MB all ok. 
The cause of the problem the following - by the end of the uploading body server already closes the connection.

How can I work around this issue and retrieve the actual response from the server?

thx, for the library, very usefull !

ovatsus:
Can you try to get a code snippet that reproduces the problem?
/cc @baronfel

baronfel:
It'll be at least 5 hours until I can look at this (traveling and whatnot) but I definitely will when I get back to my computer.

saneman1:
code snippet

    open FSharp.Data
    open System.IO

    let test () = 
        let resp = Http.Request(
                    "https://somewebdav-server/dir1/file1", //server name for example
                    httpMethod = "PUT",
                    headers = [ 
                        HttpRequestHeaders.ContentType "application/binary"
                        HttpRequestHeaders.BasicAuth "user" "pass"
                    ],
                    silentHttpErrors = true,
                    body = BinaryUpload (File.ReadAllBytes("c:\\tmp\sample.zip"))
                  )
        match resp.StatusCode with
        | 201 -> "ok" //all ok
        
       // In my case, I expect this response (and server response this), but exception raised
        | 409 -> "parent directory absent"  // no direcory

ovatsus:
Hey Chet, I didn't mean to put any kind of pressure on you, look at this whenever it's convenient, remember this is open source not work :)

baronfel:
@ovastus no worries, it's all internal pressure. I feel a tiny sense of ownership/responsibility for breaking it in the first place :)

ovatsus:
This one doesn't seem related to your change, probably was already broken.
@saneman1 , did this work in 2.3.3 and got broken on 2.4.0/2.41, or does it fail on both?

saneman1:
@ovatsus , I try only in latests version, but suppose this don't work in all.
Becouse problem in HttpWebRequest / HttpWebResponse.
[...]

PR #1112, opened 2017-12-06
Fix form body url encoding #1104

### LINKED ISSUE #1104: Large post body causes erroneous 'Invalid URI: The Uri string is too long.' error
Hey again, 

A while back I posted an issue #1077 which found that `FSharp.Data.Http.Request` would fail if the body size was too large.

This issue is to report that again the same thing is happening but now for larger body values. The biggest problem is that the error message produced is incorrect as well. When a large body is passed in the error given is `'Invalid URI: The Uri string is too long.'` when I can demonstrate that running with the same URI and just changing the body causes the failure.

This is the same example code I posted in issue #1077 except increased the size of the second test case:
```
#r @"FSharp.Data.dll"
open FSharp.Data

let bodyString = seq {for i in 0..1000 -> "x\n"} |> String.concat ""
let body = FSharp.Data.FormValues([("input",bodyString)])
FSharp.Data.Http.Request(url="http://httpbin.org/post",httpMethod="POST",body=body)
//Works fine

let bodyString2 = seq {for i in 0..20000 -> "x\n"} |> String.concat ""
let body2 = FSharp.Data.FormValues([("input",bodyString2)])
FSharp.Data.Http.Request(url="http://httpbin.org/post",httpMethod="POST",body=body2)
//throws System.UriFormatException: Invalid URI: The Uri string is too long.
```

As you can see the URI should be the same in both cases.

panesofglass:
This patch is intended to fix #1104. `Uri.EscapeDataString` fails to handle more than 32k characters, but `application/x-www-form-urlencoded` has no such restriction. While `WebUtility.UrlEncode` is [not quite sufficient](https://blogs.msdn.microsoft.com/yangxind/2006/11/08/dont-use-net-system-uri-unescapedatastring-in-url-decoding/) for correct encoding, a simple `Replace` fix can make it work.

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1513,6 +1513,12 @@ type Http private() =
 #endif
 
     static let charsetRegex = Regex("charset=([^;\s]*)", regexOptions)
+    
+    /// Correctly encodes large form data values.
+    /// See https://blogs.msdn.microsoft.com/yangxind/2006/11/08/dont-use-net-system-uri-unescapedatastring-in-url-decoding/
+    /// and https://msdn.microsoft.com/en-us/library/system.uri.escapedatastring(v=vs.110).aspx
+    static member internal UrlEncode (query:string) =
+        (WebUtility.UrlEncode query).Replace("+","%20")
 
     /// Appends the query parameters to the url, taking care of proper escaping
     static member internal AppendQueryToUrl(url:string, query) =
@@ -1592,7 +1598,7 @@ type Http private() =
                 | BinaryUpload bytes -> HttpContentTypes.Binary, (fun _ -> new MemoryStream(bytes) :> _)
                 | FormValues values ->
                     let bytes (e:Encoding) =
-                        [ for k, v in values -> Uri.EscapeDataString k + "=" + Uri.EscapeDataString v ]
+                        [ for k, v in values -> Http.UrlEncode k + "=" + Http.UrlEncode v ]
                         |> String.concat "&"
                         |> e.GetBytes
                     HttpContentTypes.FormValues, (fun e -> new MemoryStream(bytes e) :> _)
--- a/docs/content/ja/library/CsvProvider.fsx
+++ b/docs/content/ja/library/CsvProvider.fsx
@@ -405,8 +405,11 @@ airQuality.Filter(fun row -> not (Double.IsNaN row.Ozone) &&
 データセットを小さなサイズに変形した後に限定すべきです：
 *)
 
-let stocks = CsvProvider<"http://www.google.com/finance/historical?q=MSFT&output=csv", CacheRows=false>.GetSample()
-stocks.Take(10).Cache()
+let [<Literal>] ``Sacremento Real Estate`` = 
+  "http://samplecsvs.s3.amazonaws.com/Sacramentorealestatetransactions.csv"
+
+let realEstate = CsvProvider<``Sacremento Real Estate``, CacheRows=false>.GetSample()
+realEstate.Take(10).Cache()
 
 (**
 ## 関連する記事
--- a/docs/content/library/CsvProvider.fsx
+++ b/docs/content/library/CsvProvider.fsx
@@ -1,4 +1,4 @@
-(** 
+(**
 # F# Data: CSV Type Provider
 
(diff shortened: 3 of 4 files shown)


panesofglass:
cc @baronfel

baronfel:
@ovatsus I worked through this fix with @panesofglass and think it's the way to go.  Seeing some design-time failures in CI that aren't related at all though.

ovatsus:
Does this also fix #1082?

ovatsus:
The new test is failing on AppVeyor


    1) Failed : FSharp.Data.Tests.Http.32k+ bodies work
      Expected: No Exception to be thrown
      But was:  <System.Net.WebException: The remote server returned an error: (408) Request Timeout.
    Response from http://httpbin.org/post:
    <html><head><title>Request Timeout</title></head><body><p>Request Timeout.</p></body></html> ---> System.Net.WebException: The remote server returned an error: (408) Request Timeout.
       at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
       at <StartupCode$FSharp-Core>.$Control.callback@1712.Invoke(IAsyncResult ar)
       --- End of inner exception stack trace ---
       at Microsoft.FSharp.Control.AsyncBuilderImpl.commit[a](Result`1 res)
       at Microsoft.FSharp.Control.CancellationTokenOps.RunSynchronously[a](CancellationToken token, FSharpAsync`1 computation, FSharpOption`1 timeout)
       at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken)
       at FSharp.Data.Tests.Http.32k- bodies work@103.Invoke() in C:\projects\fsharp-data\tests\FSharp.Data.Tests\Http.fs:line 103
       at NUnit.Framework.Constraints.ThrowsConstraint.VoidInvocationDescriptor.Invoke()
       at NUnit.Framework.Constraints.ThrowsConstraint.ExceptionInterceptor.Intercept(Object invocation)>
    at FSharp.Data.Tests.Http.32k+ bodies work() in C:\projects\fsharp-data\tests\FSharp.Data.Tests\Http.fs:line 103
    Run Settings
        DisposeRunners: True

src/Net/Http.fs
```diff
@@ -1513,6 +1513,12 @@ type Http private() =
 #endif
 
     static let charsetRegex = Regex("charset=([^;\s]*)", regexOptions)
+    
+    /// Correctly encodes large form data values.
+    /// See https://blogs.msdn.microsoft.com/yangxind/2006/11/08/dont-use-net-system-uri-unescapedatastring-in-url-decoding/
+    /// and https://msdn.microsoft.com/en-us/library/system.uri.escapedatastring(v=vs.110).aspx
+    static member internal UrlEncode (query:string) =
+        (WebUtility.UrlEncode query).Replace("+","%20")
```
ovatsus: Is WebUtility.UrlEncode available on PCL 7 and PCL 259?

panesofglass: [`WebUtility` is available in `Profile7`](https://stackoverflow.com/a/28974461). It's available when targeting .NET 4.5. What is FSharp.Data targeting? Also, do you use `System.Web.HttpUtility` anwhere? I assume not if you are targeting PCL.

panesofglass:
I don't _think_ this is related to #1082 since that is uploading a file, not `application/x-www-form-urlencoded`.

panesofglass:
Based on NuGet.org, it seems the minimum target is .NET 4.0. Is there a definition for NET40 separate from NET45+? If so, I could use `System.Web.HttpUtility` for NET40 and `System.Net.WebUtility` for all others. Also, I saw there was [another encoding borrowed from Mono](https://github.com/fsharp/FSharp.Data/blob/1f67e15351b542b6c022d49881bcef3b3a02499e/src/Json/JsonValue.fs#L98) that you could probably do the same.

ovatsus:
We used to support .NET 4.0 and other PCL profiles, but now we only support .NET 4.5+ and PCL 7 and 259, so we just need something that's available on those 3. Maybe that code copied from Mono in JsonValue is not needed anymore

panesofglass:
Thanks, @ovatsus and @baronfel!

PR #1118, opened 2018-01-02
Support .NET Core TP loading and netstandard2.0

baronfel:
[ edited and updated by @dsyme. Joint work by @baronfel and @dsyme ]

This modernizes FSharp.Data by allowing it to be used with both .NET Core and .NET Framework/Mono-based F# tooling, including the F# compiler in the .NET SDK. 

With this work we will have FSharp.Data fully usable in both .NET Core and .NET Framework-based F# tooling.  It is also backwards compatible for use with previous F# tooling including previous editions of Visual Studio, Rider and Ionide (though if you want PCL support you have to stick to FSharp.Data 2.x)

* [x] Update all project files to .NET SDK
* [x] Multi-target `FSharp.Data.dll` to `netstandard2.0; net45`
* [x] Multi-target `FSharp.Data.DesignTime.dll` to `netstandard2.0; net45`
* [x] Rejig the nuget package layout as per [FST-1003](https://github.com/fsharp/fslang-design/blob/master/tooling/FST-1003-loading-type-provider-design-time-components.md) 
* [x] Remove support for PCLs of FSharp.Data.dll
* [x] Test on both `netcoreapp2.0` and `net46` using the .NET SDK
* [x] Remove use of old version of SourceLink 1.x
* [x] Bump to version 3.0.0-beta
* [x] Test on `net46` using Mono `msbuild` toolchain and F# compiler 
* [x] Test on `net46` using .NET Framework `msbuild` toolchain and F# compiler (at the moment all of Linux/OSX/Windows testing is with .NET SDK `dotnet` toolchain)
* [x] Set USE_MSBUILD=1 as part of appveyor matrix
* [x] Run tests when USE_MSBUILD=1 set
* [x] Opening FSharp.Data.sln and building inside VS2017 just works
* [x] Disable failing test that requires compiler to run with .NET Framework, unless USE_MSBUILD is set
* [x] AppVeyor CI passes
* [x] Document workarounds for missing encodings on .NET core (in RELEASE_NOTES.md)
* [ ] Travis CI passes
* [x] Opening FSharp.Data.Tests.sln and running tests inside VS2017 just works

To look at later (after this PR I think):
* [ ] The documentation tests only run on Mono but feel flaky.  For example, google started blocking some google finance requests when trying to run them on my Windows machine.
* [ ] I disabled the World Bank signature tests on the local machine (they are already disabled on the build server).  Updating these should be orthogonal
* [ ] The PR removed the use of `paket.references` in favour of explicit `PackageReference` because of https://github.com/fsprojects/Paket/issues/3091.  

Notes:
* FSharp.Data multi-targets to `net45; netstandard2.0`. We could add more, e.g. netstandard1.6, though there's not much need)
* FSharp.Data.DesignTime multi-targets to `net45; netstandard2.0`. We don't need to add more since F# tooling always assumes one of these
* Use .NET SDK version 2.1.100-preview-007363 to test this
* One fix was needed to the TPSDK, that is incorporated

### OLD NOTES

A new fake build task builds the netcore stuff and the nuspec has been updated with the appropriate version constraints.

Sadly it appears we have to maintain this two-solution approach because the PCL versions of FSharp.Core have been dropped as of somewhere around FSharp.Core version 4.1.17 or 18, so we need an old-project-file-style build to handle building PCLs. 

Bonus points, and something I tried very hard to do, was get some subset of the tests building on .Net standard 2.0 as well.  I ran into problems here because of Nancy dependencies not playing nicely, so perhaps an enhancement to switch the test server over to Suave?


baronfel:
image

Package structure after this work^

dsyme:
@baronfel I'm going to make a push to get FSharp.Data working with the previews of the latest .NET SDKs and able to target netstandard2.0 and netcoreapp2.0.  I'll take a look at this PR as part of this

> Sadly it appears we have to maintain this two-solution approach because the PCL versions of FSharp.Core have been dropped as of somewhere around FSharp.Core version 4.1.17 or 18, so we need an old-project-file-style build to handle building PCLs.

I think we should just drop the PCLs now and declare a major version update to FSharp.Data.  People building PCLs should continue to use the old version.

baronfel:
I'd be in favor of dropping PCL, I just wish we had some data to put behind making that decision. The main use case for PCLs used to be mobile development, right? And those environments now have runtimes that are net standard 2.0 compatible?

dsyme:
@baronfel Yes, the Xamarin runtimes all now support .NET Standard 2.0.  Support for PCLs is more or less being deprecated across the whole ecosystem.  

That said, .NET Standard 2.0 is not the specific nirvana for mobile - for the shared code in mobile apps you actually want multi-targeting, where the shared code targets both MonoDroid (Android) and MonoTouch (iOS)

ovatsus:
Does this mean that only F# 4.1 will be supported, and support for 3.1 and 4.0 will be dropped?

dsyme:
> Does this mean that only F# 4.1 will be supported, and support for 3.1 and 4.0 will be dropped?

@ovatsus I don't think that's necessary. 

Note that it's not about which F# language version but which FSharp.Core versions the TPRTC(s) are compatible with and which F# tooling versions/deployments the TPDTC(s) will load into.  It should be possible to maintain compat though perhaps tricky to test 

I'm making some adjustments to this PR and I'll draw up a matrix when I submit, thanks.

dsyme:
@baronfel I have submitted work directly to this PR to do the following

* Support FST-1003 to allow the FSharp.Data TPDTCs type provider to be used with .NET Core-based tooling.  
* Use .NET SDK version 2.1.100-preview-007363 to test this
* Remove support for PCLs
* Add a 
* Bump to version 3.0.0-beta
*

dsyme:
Ready to review

baronfel:
This is amazing, and I'll take a deep look at it tonight!

dsyme:
@baronfel Notes here: https://github.com/fsprojects/FSharp.TypeProviders.SDK/blob/master/README.md#updating-a-type-provider-to-be-suitable-for-use-with-the-net-sdk

baronfel:
I note that Sourcelinking has been removed. There's an updated, dotnet-sdk compatible version of it that hadn't been intregrated into this repo because we hadn't gone to the dotnet cli. I could try to integrate that as well.

baronfel:
The updated docs are fantastic! Also very nice to have worked examples in the BasicProvider.

thinkbeforecoding:
Is there still something blocking this PR from being merged ?

ovatsus:
I'd like to review it before merging to master since it's such a big change and we're dropping support for some versions, haven't had the time yet, we'll try to get to it soon

baronfel:
Thanks to @ctaggart I was able to get sourcelink reconfigured for the dotnet sdk. Dunno why the build didn't kick off on this side, though. It was green on my PR branch as seen here: https://github.com/baronfel/FSharp.Data/pull/1

thinkbeforecoding:
This seems to be just a conflict with release notes 😋

appveyor.yml
```diff
@@ -4,7 +4,6 @@ build_script:
   - cmd: build.cmd CleanInternetCaches
   - cmd: build.cmd AppVeyorBuildVersion
   - cmd: build.cmd All
-  - cmd: build.cmd SourceLink
```
ovatsus: @baronfel , didn't you fix SourceLink on the last commit?

nuget/FSharp.Data.nuspec
```diff
@@ -16,43 +16,40 @@
...
+      <group targetFramework="net45">
         <dependency id="FSharp.Core" version="4.0.0.1"/>
       </group>
-      <group targetFramework="portable-net45+netcore45">
-        <dependency id="Zlib.Portable" version="1.11.0" />
-      </group>
-      <group targetFramework="portable-net45+netcore45+wpa81+wp8">
-        <dependency id="Zlib.Portable" version="1.11.0" />
-      </group>
       <group targetFramework="net40" />
```
ovatsus: net40 is no longer supported, can you remove this?

README.md
```diff
@@ -22,12 +22,13 @@ You can see the version history [here](RELEASE_NOTES.md).
...
 - FSharp.Core 4.4.0.0 (F# 4.0)
 - FSharp.Core 4.4.1.0 (F# 4.1)
+- FSharp.Core 4.4.3.0 (F# 4.1 update)
 
 ## Supported platforms
 
-- .NET Framework 4.5
-- PCL profile 7 (.NET Framework 4.5, Windows 8)
-- PCL profile 259 (.NET Framework 4.5, Windows 8, Windows Phone 8.1, Windows Phone Silverlight 8)
+- .NET Framework 4.x
```
ovatsus: 4.0 is not supported, only 4.5 and 4.6, right? So can you change this to 4.5+

README.md
```diff
@@ -22,12 +22,13 @@ You can see the version history [here](RELEASE_NOTES.md).
...
+- .NET Core App 2.x
+- .NET Standard 2.0
```
ovatsus: Forgive my ignorance, but what's the difference between .NET Core App 2.x and .NET Standard 2.0?

nuget/FSharp.Data.nuspec
```diff
@@ -16,43 +16,40 @@
...
+
+    <!-- Place the TPDTC in the legacy location alongside FSharp.Data --> 
+    <file src="..\src\FSharp.Data.DesignTime\bin\Release\net45\FSharp.Data.DesignTime.dll" target="lib/net45" />
+    <file src="..\src\FSharp.Data.DesignTime\bin\Release\net45\FSharp.Data.DesignTime.pdb" target="lib/net45" />
+
+    <!-- Place the TPDTC in the future-proof location --> 
+    <file src="..\src\FSharp.Data.DesignTime\bin\Release\net45\FSharp.Data.DesignTime.dll" target="typeproviders/fsharp41/net45" />
+    <file src="..\src\FSharp.Data.DesignTime\bin\Release\net45\FSharp.Data.DesignTime.pdb" target="typeproviders/fsharp41/net45" />
+    <file src="..\src\FSharp.Data.DesignTime\bin\Release\netstandard2.0\FSharp.Data.DesignTime.dll" target="typeproviders/fsharp41/netstandard2.0" />
+    <file src="..\src\FSharp.Data.DesignTime\bin\Release\netstandard2.0\FSharp.Data.DesignTime.pdb" target="typeproviders/fsharp41/netstandard2.0" />
```
ovatsus: I'm reading a bunch of stuff about this, but still didn't really understand why the design time dll going forward cannot be in the same dir as the runtime dll but has to be on a typeproviders subdirectory

README.md
```diff
@@ -22,12 +22,13 @@ You can see the version history [here](RELEASE_NOTES.md).
 - FSharp.Core 4.3.1.0 (F# 3.1)
 - FSharp.Core 4.4.0.0 (F# 4.0)
 - FSharp.Core 4.4.1.0 (F# 4.1)
+- FSharp.Core 4.4.3.0 (F# 4.1 update)
```
ovatsus: F# version has got a bit confusing with all the changes, and saying "F# 4.1 update" doesn't help that much. Would you mind changing this to something like this:

- FSharp.Core 4.3.1.0 (F# 3.1 SDK / Visual Studio 2013)
- FSharp.Core 4.4.0.0 (F# 4.0 SDK / Visual Studio 2015)
- FSharp.Core 4.4.1.0 (F# 4.1 SDK / Visual Studio 2017)
- FSharp.Core 4.4.3.0 (F# 10.1 SDK / Visual Studio 2017 15.6)

ovatsus: Also, if I understood correctly, you need VS2017 15.6 to be able to build FSharp.Data now, so can you update the line in line 13 to say Visual Studio 2017 15.6+ ?

tests/FSharp.Data.Tests/WorldBankProvider.fs
```diff
@@ -1,7 +1,12 @@
-﻿#if INTERACTIVE
-#r "../../bin/FSharp.Data.dll"
+﻿#if NETSTANDARD2_0
+module FSharp.Data.Tests.WorldBankProvider
+do ()
```
ovatsus: why id this do() here needed?

ovatsus: it seems to be the only reason we need the #if NETSTANDARD2_0 at all on all these test files

tests/FSharp.Data.Tests/CsvProvider.fs
```diff
@@ -299,19 +303,23 @@ let ``Csv without sample``() =
...
 let ``Uses UTF8 for sample file when encoding not specified``() =
     let utf8 = UTF8.GetSample()
     let row2 = utf8.Rows |> Seq.skip 1 |> Seq.head
     row2 |> should equal (2, "NaN (�񐔒l)")
 
+#if !NETCOREAPP2_0 // "No data is available for encoding 932. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method."
+type CP932 = CsvProvider<"Data/cp932.csv", Culture = "ja-JP", Encoding = "932", HasHeaders = true, MissingValues = "NaN (非数値)">
+
 [<Test>]
+[<Ignore("Disabled for now - appears to fail on .NET Core 2.0 - No data is available for encoding 932 - and also has issues on .NET Framework 4.6. ")>]
```
ovatsus: Can we keep it running on .Net Framework?

tests/FSharp.Data.Tests/HttpIntegrationTests.fs
```diff
@@ -82,8 +83,9 @@ let ``when called on a non-existant page returns 404`` () =
     Http.Request("http://localhost:1235/TestServer/NoPage", silentHttpErrors=true).StatusCode |> should equal 404
 
 [<Test>]
-[<Platform("Net")>]
+//[<Platform("Net")>]
 let ``all of the manually-set request headers get sent to the server`` ()=
+  if not runningOnMono then
```
ovatsus: Doens't Platform("Net") work anymore?

tests/FSharp.Data.DesignTime.Tests/DocumentationTests.fs
```diff
@@ -71,10 +71,12 @@ for file in docFiles do
 
 [<Test>]
 [<TestCaseSource "docFiles">]
-[<Platform("Mono")>]
+//[<Platform("Mono")>]
```
ovatsus: Is the platform attribute not supported anymore?

ovatsus:
I may have gotten this wrong, but doesn't PackageReference always use NuGet? Then in that case what is paket still used for? On the early days one of the big advantages of paket was not having version numbers in the package folders, which meant not breaking the fsx scripts everytime a package was a upgraded. Is that now fixed on NuGet 3.0? I'm really not been following the latest developments here. /cc @forki

src/FSharp.Data.DesignTime/FSharp.Data.DesignTime.fsproj
```diff
@@ -0,0 +1,72 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\netfx.props" />
+  <PropertyGroup>
+    <TargetFrameworks>net45; netstandard2.0</TargetFrameworks>
+    <DefineConstants>NO_GENERATIVE</DefineConstants>
```
ovatsus: Is this still needed with the latest TPSDK?

README.md
ovatsus: Actually looking at the packagereferences, for .net standard we support FSharp.Core 4.4.1.0 (fsharp.core 4.2.3 nuget package) and 4.4.3.0, maybe join this section with the supported platform one so we have the breakdown of supported fsharp.core versions per platform?

src/FSharp.Data.DesignTime/FSharp.Data.DesignTime.fsproj
```diff
@@ -0,0 +1,72 @@
...
+    <Compile Include="..\Html\HtmlRuntime.fs" />
+    <Compile Include="..\Html\HtmlGenerator.fs" />
+    <Compile Include="..\Html\HtmlProvider.fs" />
+    <Compile Include="..\TypeProviderInstantiation.fs" />
+    <Compile Include="..\AssemblyInfo.DesignTime.fs" />
+    <None Include="..\Test.fsx" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="FSharp.Core" Version="4.0.0.1" Condition="'$(TargetFramework)' == 'net45'" />
+    <PackageReference Include="FSharp.Core" Version="4.2.3" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```
ovatsus: Is this right? Nuget package 4.2.3 doesn't say it supports netstandrd 2.0. Maybe for .net standard we would use 4.3.4 and only support FSharp.Core 4.4.3.0 and not 4.4.1.0 to reduce the test matrix, since it was never supported before anyway?

src/Net/Http.fs
```diff
@@ -1145,6 +1145,7 @@ module private HttpHelpers =
         }
 
     let writeBody (req:HttpWebRequest) (data: Stream) =
+#if FX_NO_LOCAL_FILESYSTEM
```
ovatsus: This seems unrelated to not having local filesystem, should a separate FX_xxx be added?

ovatsus:
Why do we now need the Program.fs for the tests?

paket.dependencies
```diff
@@ -1,22 +1,50 @@
...
+    nuget FSharp.Charting
+    nuget FSharp.Formatting
+    nuget Nancy.Hosting.Self
+
+# For testing against FSharp.Core for F# 3.1
+group fs31
+    source https://api.nuget.org/v3/index.json
+    nuget FSharp.Core ~> 3.1.0
+
+# For testing against FSharp.Core for F# 4.0
```
ovatsus: I see these groups defined here, but I can't see what makes the tests run against all these different F# versions

RELEASE_NOTES.md
```diff
@@ -1,3 +1,8 @@
+#### 3.0.0-beta - February 27 2018
+* Drop SourceLink support
```
ovatsus: No longer true, right?

appveyor.yml
ovatsus: Or does it always run now without a separate command?

ovatsus:
So what makes this require dropping support the PCLs? I actually have some projects that use PCLs that aren't migratable to .net standard, so if we could still make them work, that would be great. Wouldn't just having multiple FSharp.Data csproj's still work like before?

If we're removing them, however, then there's actually a lot of stuff that can be cleaned up in Http.fs and IO.s

ovatsus:
Got this when building this branch locally:

      1) System.Exception: dotnet command failed on sourcelink test bin/Release/netstandard2.0/FSharp.Data.DesignTime.pdb
       at Fake.DotNetCli.RunCommand@78-3.Invoke(String message) in D:\code\fake\src\app\FakeLib\DotNetCLIHelper.fs:line 78
       at Fake.DotNetCli.RunCommand(FSharpFunc`2 setCommandParams, String args) in D:\code\fake\src\app\FakeLib\DotNetCLIHelper.fs:line 78
       at FSI_0005.Build.testSourcelink@124.Invoke(String proj)
       at FSI_0005.Build.clo@110-10.Invoke(Unit unitVar0)
       at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in D:\code\fake\src\app\FakeLib\TargetHelper.fs:line 626

dsyme:
> So what makes this require dropping support the PCLs? I actually have some projects that use PCLs that aren't migratable to .net standard, so if we could still make them work, that would be great. Wouldn't just having multiple FSharp.Data csproj's still work like before?

I really think we need to drop PCL support 

1. you can't build PCL projects with .NET SDK tooling
2. projects that require this (Windows 8 Phone?) can continue to use an existing version of the type provider
3. the complexity it introduces to support and test them is mind-blowing

That said - yes, technically everything will continue to work if you have a PCL runtime DLL.

README.md
dsyme: .NET Standard 2.0 is a big and large common library subset supported supported by .NET Framework 4.6+, Mono (including Xamarin Android/iOS) and .NET Core. It has a normal reflection model and has pretty much everything you actually need to build most libraries, and looks and feels like normal .NET 4.x programming for nearly all purposes.  All of FSharp.Data compiles without any `#if` when targeting .NET Standard 2.0.  It's basically what PCLs should always have been.

.NET Core App 2.0 (netcoreapp2.0) is the name of the programming model for a typical .NET Core application.  The programming model is basically the same as .NET Standard 2.0 but in theory you could have access to extra stuff that is not supported by .NET Standard (e.g. cross-platform ways of accessing .NET Core specifics about the .NET Core GC or something).

README.md
dsyme: Done thanks

README.md
dsyme: Fixed, thanks. Also clarified the design-time environments supported

paket.dependencies
dsyme: Sorry - it comes from here: https://github.com/fsprojects/FSharp.TypeProviders.SDK/blob/master/src/ProvidedTypesTesting.fs#L681

I've added that as a comment
```
. The TPSDK testing helpers like these to be placed in package groups
# with these names, see https://github.com/fsprojects/FSharp.TypeProviders.SDK/blob/0c0c9d11135039821a380f60ec21c701354e8cdb/src/ProvidedTypesTesting.fs#L680
```

src/FSharp.Data.DesignTime/FSharp.Data.DesignTime.fsproj
dsyme: Since this is an erasing type provider, using this define makes the design-time DLL smaller.  

But yes, we could now optionally make this a generative type provider.  That would be separate thing and I think would have to be revealed as a separately named TP, since whether a TP is generative or not is determined before the static parameters are applied:
```fsharp
type MyXml = XmlProviderGenerative<"foo.xml">
```
Besides I believe it would be easy (it may stilll uncover bugs in generative TP support, though that's pretty solid - Swagger has ironed out the issues).

src/FSharp.Data.DesignTime/FSharp.Data.DesignTime.fsproj
dsyme: @ovatsus Yes, that is better 4.2.3 supports `.netstandard1.6` which is compatible with `.netstandard2.0`.  But `4.3.4` adds an explicit empty dependency node which removes transitive dependency on all the micro packages that existed in the world of .NET Standard 1.x

tests/FSharp.Data.DesignTime.Tests/DocumentationTests.fs
dsyme: I can't recall the problem I had with this but yes, I concluded it wasn't supported in .net standard .NUnit as yet.  I think it's an omission in the .Net standard NUnit DLL

tests/FSharp.Data.Tests/CsvProvider.fs
dsyme: I couldn't work out what's going on here, it's a really, really odd problem.  My impression is that the test is flakey in some way even on .NET Framework.  I looked again and again and can't see anything that I changed that may affect this - except perhaps moving from .NET Framework 4.5 to 4.6.  I'll try again.

tests/FSharp.Data.Tests/HttpIntegrationTests.fs
dsyme: See above on .NET Standard NUnit

tests/FSharp.Data.Tests/WorldBankProvider.fs
dsyme: Great catch, I think that can be removed, trying it now

dsyme:
> I may have gotten this wrong, but doesn't PackageReference always use NuGet? Then in that case what is paket still used for? On the early days one of the big advantages of paket was not having version numbers in the package folders, which meant not breaking the fsx scripts everytime a package was a upgraded. Is that now fixed on NuGet 3.0? I'm really not been following the latest developments here. /cc @forki

Yes, I considered this. The advantages with .NET SDK projects seem to be
* make sure we have a consistent view of the references **and their transitive dependencies**
* allow package groups to download stuff for testing

But I agree the "why paket" question feels different in the era of .NET SDK projects and the arguments are more subtle.  Let's consider this separately.

dsyme:
Re PCLs:
> If we're removing them, however, then there's actually a lot of stuff that can be cleaned up in Http.fs and IO.s

Yes, loads of cleanup can be done.  Let's do it in a separate PR though

appveyor.yml
dsyme: It was removed separately to this PR, this flushed some more stuff from build.fsx

RELEASE_NOTES.md
baronfel: Correct. I added it back in, and it seamlessly happens during the build step.

appveyor.yml
baronfel: Yes, though I did add it back in using the new clitool and msbuild tasks that @ctaggart helped me get working. It now happens transparently during the build, due to the presence of the `SourceLinkCreate` msbuild property.

There's a separate testing step that happens to verify the sourcelinking occurred correctly.

dsyme:
With regard to this failing test: https://github.com/fsharp/FSharp.Data/pull/1118/files/81222c41428a6ccc4e8b748fbf85f276fdab3f91#diff-a5c5684880691154645529d05f1d3780R317

This is tricky. The encoding `932` is apparently not supported by .NET Core, even when running on Windows, for `System.Text.Encoding.GetEncoding(932)` we get:
```
Unhandled Exception: System.NotSupportedException: No data is available for encoding 932. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method.
```

I think .NET Core just doesn't support this encoding.

This means that when F# design-time tooling runs using .NET Core, FSharp.Data design-time tooling can't be used with files that have this encoding.

I believe the correct thing to do is 

1. Move this test to FSharp.Data.DesignTime

2. Give a design-time error when `System.Text.Encoding.GetEncoding` raises an exception, hinting that the workaround from https://github.com/Microsoft/visualfsharp/issues/3303 may need to be applied.

dsyme:
@ovatsus Just to mention that this [catch-all exception handler] (https://github.com/fsharp/FSharp.Data/blob/master/src/CommonProviderImplementation/Helpers.fs#L245) is really covering up a lot of potential problems :)

For example, on CoreCLR we were failing GetEncoding(932) but the exception was being hidden.  We were then just parsing the filename as some random bit of one-column CSV text....  

Then I tried to fix that and we were then failing a DLL load of a helper DLL....

dsyme:
@ovatsus I've finally found the cause of the bug with codepages, but want to ask you a few things..

The bug is that integer code pages like 932 weren't working with .NET Core design-time tooling, giving very obscure results. This is because .NET Core chops out 1MB of codepage tables out of the minimal profile and you have to opt-in to support for those

One way to fix this is to for both FSharp.Data.dll and FSharp.Data.DesignTime.dll to take a dependency on package "System.Text.Encoding.CodePages", which is a .NET Standard package to solve this problem for .NET Core apps.  However it means that using FSharp.Data will always induce a dependency on System.Text.Encoding.CodePages, which is about 1MB.

I'm not sure what to do about this.  The design-time support could have this dependency, but it feels bad if FSharp.Data.dll has it too.  But without it it is difficult to ensure that `GetSample()` would use the same codepage as the one provided at design-time.

TBH it would be much, much simpler to drop integer codepage support from FSharp.Data altogether, or require the use of the [workaround](https://github.com/Microsoft/visualfsharp/issues/3303) when using .NET Core-based tooling.

ovatsus:
Yes, let's drop it from the .NET Core version, I'd rather not have that extra dependency, people can still use the .Net Framework version. If System.Text.Encoding.CodePages.dll is present, will it just work? If so, we can just add that to the documentation, explaining that it need to be copied both to the runtime and TP folders

dsyme:
@ovatsus 

> Yes, let's drop it from the .NET Core version, I'd rather not have that extra dependency, people can still use the .Net Framework version. If System.Text.Encoding.CodePages.dll is present, will it just work? If so, we can just add that to the documentation, explaining that it need to be copied both to the runtime and TP folders

Just to clarify - "dropping from the .NET Core version" means that integer code pages will not work with the F# compiler that comes as part of the .NET SDK, which executes using .NET Core and which more and more people are using.  You have to install the .NET Framework or Mono and use the workaround mentioned above whenever using the dotnet SDK.

The feature won't work when `System.Text.Encoding.CodePages.dll` is present: it is a design-time feature and we have to either support it at design-time or not: we would need to include that DLL as part of the design-time components in `typeproviders\...` (BTW one of the reasons we're moving the design-time components to a separate directory is to clarify this kind of thing).

ovatsus:
But wouldn't it work work if you copy the dll both to typeprovider\... and to the runtime folder?

baronfel:
I've been digging a bit more into the Sourcelink side of this and I believe @ovatsus that what you're seeing is related to mismatched line endings. The `sourcelink test` step works for me when testing this branch on my OSX machine, but fails as you describe on my windows machine.  I'll continue researching that.

tests/FSharp.Data.Tests/WorldBankProvider.fs
baronfel: I just went through all these test files and removed the netstandard2_0 if conditions. All tests still run and pass for me locally.

src/FSharp.Data/FSharp.Data.fsproj
```diff
@@ -0,0 +1,78 @@
...
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
+
+  <!-- We build BOTH the design-time DLLs needed to support the use of the runtime DLL in different host tooling -->
+  <!-- In the future it will be possible to just build a single netstandard2.0 design-time DLL, but for fsharp41 host tooling  -->
+  <!-- that is not yet possible because of https://github.com/Microsoft/visualfsharp/issues/3309. Also for legacy loading -->
+  <!-- we still need a .NET 4.5 design-time DLL to place alongside the runtime DLL -->
+
+  <!-- We place BOTH the design-time DLLs in well-known locations with resepct to the runtime DLL -->
+  <!-- This enables any runtime DLL to be used with any host tooling running in either netcoreapp2.0+ or net461+ -->
+  <Target Name="AfterBuild">
```
baronfel: are these copies still needed since we are doing explicit layout of files in the nuspec?  The layout there seems to mirror this quite closely.

baronfel:
Doubly interesting that everything is working with regards to sourcelink in the CI environments.

baronfel:
@ctaggart here's the test output from the build on my windows machine. Note that this is on the preview version 2.1.100-preview-007363 of the SDK:

```
C:\Users\chusk\AppData\Local\dotnetcore\dotnet.exe sourcelink test bin/Release/netstandard2.0/FSharp.Data.DesignTime.pdb
43 Documents without URLs:
26549200fd6b4bfbea2f50f5896b5cd5 md5 fsharp D:\code\fsharp.data\src\Net\Http.fs
21afc3a672bb73027755224b174543c7 md5 fsharp D:\code\fsharp.data\src\CommonRuntime\TextConversions.fs
a7bc4c4d13cce38931bfec2e9719a0f6 md5 fsharp D:\code\fsharp.data\src\CommonRuntime\TextRuntime.fs
763c766ff5b4ffb7802e8207e0fa7932 md5 fsharp D:\code\fsharp.data\src\CommonRuntime\StructuralTypes.fs
48da8131935c80e129d839d1695f1cdf md5 fsharp D:\code\fsharp.data\src\CommonProviderImplementation\Helpers.fs
081e042f6473c4be4c0eec63f9b1f8f1 md5 fsharp D:\code\fsharp.data\src\Json\JsonValue.fs
0898ecff4e57463d7f7e445dac4dfc9f md5 fsharp D:\code\fsharp.data\src\Json\JsonConversions.fs
3f8260b7b1b2fce0ec4f870bfaeb568d md5 fsharp D:\code\fsharp.data\src\Json\JsonExtensions.fs
f2878909063c8a66b1aa37ce226a3ffd md5 fsharp D:\code\fsharp.data\src\Json\JsonRuntime.fs
0c0d6b33265312d1a1eb1887ec1672f8 md5 fsharp D:\code\fsharp.data\src\Json\JsonGenerator.fs
5d12bcb6530a1dd9d5a0a947a4d9c35a md5 fsharp D:\code\fsharp.data\src\Json\JsonProvider.fs
e3201464739ddea79c2332d61e7f13bf md5 fsharp D:\code\fsharp.data\src\Xml\XmlRuntime.fs
8c5a1dded8a786f0f937a4d82079f947 md5 fsharp D:\code\fsharp.data\src\Xml\XmlGenerator.fs
f8edd6abb376b7205e90486e65038df1 md5 fsharp D:\code\fsharp.data\src\Xml\XmlProvider.fs
828c8a712ed23f8831fdf25fcaca8428 md5 fsharp D:\code\fsharp.data\src\Csv\CsvRuntime.fs
75853e375d0b4b71d7336441180c784a md5 fsharp D:\code\fsharp.data\src\Csv\CsvFile.fs
10314cc8e06bb65376953a6f54611916 md5 fsharp D:\code\fsharp.data\src\Csv\CsvExtensions.fs
f3c5f0d201ae6309cb85291fc793d01a md5 fsharp D:\code\fsharp.data\src\Csv\CsvProvider.fs
8cfefd301be6e50f9823d1db8642866f md5 fsharp D:\code\fsharp.data\src\WorldBank\WorldBankRuntime.fs
e2cf8415924941c19542be6936b7c658 md5 fsharp D:\code\fsharp.data\src\WorldBank\WorldBankProvider.fs
c420bff9a7d0627d90aa934c29b8c1f9 md5 fsharp D:\code\fsharp.data\src\Html\HtmlParser.fs
59b2405e23bb6a1fbaa63573e0779d3e md5 fsharp D:\code\fsharp.data\src\Html\HtmlOperations.fs
6666548d51ba339257efac471e1d1881 md5 fsharp D:\code\fsharp.data\src\Html\HtmlRuntime.fs
fe0d1e13348c699bd9ab363c18b363f2 md5 fsharp D:\code\fsharp.data\src\Html\HtmlProvider.fs
75ef69fbcb0b28458d35510ead4e8537 md5 fsharp D:\code\fsharp.data\src\TypeProviderInstantiation.fs
1723e8043b9e971e2214f5765262fec6 md5 fsharp D:\code\fsharp.data\src\Html\HtmlGenerator.fs
f810e0fd8ebf96258752b0a72ee7988c md5 fsharp D:\code\fsharp.data\src\Html\HtmlInference.fs
257c1ffcb58d93c602dbd2884a22e56c md5 fsharp D:\code\fsharp.data\src\Html\HtmlCssSelectors.fs
626f82bb449c6c8c07c941e49105074d md5 fsharp D:\code\fsharp.data\src\Html\HtmlCharRefs.fs
d23f8012c77bb11eacef3fea77adf9a7 md5 fsharp D:\code\fsharp.data\src\Csv\CsvGenerator.fs
8e99eacc478438f6394229b18d62a0fa md5 fsharp D:\code\fsharp.data\src\Csv\CsvInference.fs
54be0e587f7e242fea500a4081d2ff7c md5 fsharp D:\code\fsharp.data\src\Xml\XmlInference.fs
60b3f4b367a8041f3e1db72f7ba1706e md5 fsharp D:\code\fsharp.data\src\Json\JsonConversionsGenerator.fs
3227d4ab40b25db6e644508c04b6cbe1 md5 fsharp D:\code\fsharp.data\src\Json\JsonInference.fs
97e4058d6d8312f83d2e01124d10d3dc md5 fsharp D:\code\fsharp.data\src\CommonProviderImplementation\ConversionsGenerator.fs
d1c82bc8e36a09afe61ae874bb55eb13 md5 fsharp D:\code\fsharp.data\src\CommonProviderImplementation\AssemblyResolver.fs
65e744c4563725e2032f32c5538471d2 md5 fsharp D:\code\fsharp.data\src\CommonProviderImplementation\QuotationBuilder.fs
b1715d94b1eb3fffadad6531c2f83584 md5 fsharp D:\code\fsharp.data\src\CommonRuntime\StructuralInference.fs
12a47fdd3484dafbabacb28d3185516d md5 fsharp D:\code\fsharp.data\src\CommonRuntime\NameUtils.fs
824ae16cccd73fc16d28412cd21075e4 md5 fsharp D:\code\fsharp.data\src\CommonRuntime\Pluralizer.fs
2e028322daa02d448a80d36a6c7e341d md5 fsharp D:\code\fsharp.data\src\CommonRuntime\Caching.fs
fcafdcafbbfbe2f265d197c692225c05 md5 fsharp D:\code\fsharp.data\src\CommonRuntime\IO.fs
ac7b324806b5b42052d4e85bfd8d849c md5 fsharp D:\code\fsharp.data\src\Net\UriUtils.fs
sourcelink test failed
```

if I dig into the sourcelink json file, here's its contents:
```
{"documents":{"D:\\Code\\fsharp.data\\*":"https://raw.githubusercontent.com/baronfel/FSharp.Data/cf5f2b05cc1b25eae18209d776a8f910fe05ad49/*"}}
```

One immediate discrepancy that I see is that the paths expected in the test output above have a lowercase 'c' in the path `D:\code`, whereas the documents pathspec in the sourcelink json has a capital 'C'.

baronfel:
@ctaggart can confirm that it's a path casing discrepancy. When I changed directory to match the casing stored in sourcelink.json the tests passed.  I think we need to take a look at how the root directory is written [here](https://github.com/ctaggart/SourceLink/blob/1f46401d5590a3d9db9958e4bcc7b0c9127b3c61/SourceLink.Create.CommandLine/CreateTask.cs#L56) and how it is matched using regex [here](https://github.com/ctaggart/SourceLink/blob/1f46401d5590a3d9db9958e4bcc7b0c9127b3c61/dotnet-sourcelink/Program.cs#L450)

ovatsus:
I did a `build all` on the latest changes on this branch and didn't get any source link error

tests/FSharp.Data.Tests/JsonProvider.fs
```diff
@@ -480,13 +480,13 @@ let ``Can parse UTC dates``() =
     dates.UtcTime.ToUniversalTime() |> should equal (new DateTime(1997, 7, 16, 19, 50, 30, 0)) 
 
 [<Test>]
-[<SetCulture("zh-CN")>]
+[<Culture("zh-CN")>]
```
ovatsus: This doesn't have the same effect. The previous attribute would switch to that culture when running the test and then switch back at the end, the new one just ignores the test unless you're running on that culture, which is not what we want. Output when running the tests on this branch:

      When running the tests I get this:
      Not Run  Can parse ISO 8601 dates in the correct culture
      Error Message:
       Only supported under culture zh-CN

      Not Run  Can parse ISO 8601 dates in the specified culture
      Error Message:
       Only supported under culture pt-PT

      Not Run  Can parse dates in different cultures
      Error Message:
       Only supported under culture pt-PT

      Not Run  Can parse decimals in different cultures
      Error Message:
       Only supported under culture pt-PT

      Not Run  Can parse document with iso date in local culture
      Error Message:
       Only supported under culture zh-CN

      Not Run  Can parse floats in different cultures
      Error Message:
       Only supported under culture pt-PT

guguer:
@baronfel can you give me permissions to your pull request branch so I can push further fixes? https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

baronfel:
@guguer I think that's been done, I see that box checked on this PR.

Would you mind editing the checklist at the top with remaining things you'd like to see? The discussion has gotten a bit longwinded, and a list would help me nail down the final items.

guguer:
I'm getting this:

image

image

baronfel:
Odd! I've checked the box here:
image

I invited you as a collaborator to my fork, so perhaps that will grant you access?

guguer:
Thanks, it worked now

ovatsus:
I think it didn't work before because I messed up the git commands

ovatsus:
The is one problem with SourceLink being integrated, though, when you build locally without having pushed it always fails :/ @baronfel , can we change it so it still runs SourceLink as part of the release task and not the build task? Or maybe leave a is and output the error at the end but still carry on with the test tasks when SourceLink fails?

ovatsus:
I'm able to build and tests everything fine in the command line (even though CI is now failing :/), but when opening the tests solution in VS, I get a bunch of errors like this:

    C:\Git\FSharp.Data\tests\FSharp.Data.Tests\FSharp.Data.Tests.fsproj : error NU1105: Unable to find project information for 'C:\Git\FSharp.Data\src\FSharp.Data\FSharp.Data.fsproj'. The project file may be invalid or missing targets required for restore.

If I manually build it still works fine, but there's something weird going on, I also get warning on solution explorer:

image

baronfel:
@ovatsus agreed on making sourcelink a release step not a build step. Working on that right now :)

build.fsx
```diff
@@ -227,12 +225,23 @@ Target "ReleaseBinaries" <| fun () ->
...
+    |> Seq.collect (fun fw -> buildProjs |> Seq.map (testSourcelink fw))
+    |> Seq.iter id
+
 Target "Release" DoNothing
 
 "CleanDocs" ==> "GenerateDocs" ==> "ReleaseDocs"
 "ReleaseDocs" ==> "Release"
 "ReleaseBinaries" ==> "Release"
 "NuGet" ==> "Release"
+"TestSourcelink" ==> "Release"
```
ovatsus: Update help bellow to mention this new target

baronfel: Thanks for the reminder!

baronfel:
@ovatsus after updating to VS 15.6.4 I have the same errors you mention.

ovatsus:
And master also stopped building  the tests after the update :(

      1) Building C:\Git\FSharp.Data\FSharp.Data.Tests.sln failed with exitcode 1.
      2) : C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(186,5): Your project is not referencing the ".NETFramework,Version=v4.5" framework. Add a reference to ".NETFramework,Version=v4.5" in the "frameworks" section of your project.json, and then re-run NuGet restore.
      3) : C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(186,5): Your project is not referencing the ".NETFramework,Version=v4.5" framework. Add a reference to ".NETFramework,Version=v4.5" in the "frameworks" section of your project.json, and then re-run NuGet restore.
      4) : C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(186,5): Your project is not referencing the ".NETFramework,Version=v4.5" framework. Add a reference to ".NETFramework,Version=v4.5" in the "frameworks" section of your project.json, and then re-run NuGet restore.

baronfel:
Similar stuff on my end, but I got a bit further by explicitly adding FSharp.Data.fsproj and FSharp.Data.DesignTime.fsproj to the tests solution.  [This StackOverflow question](https://github.com/NuGet/Home/issues/5350) suggests it could be a casing error somewhere?

ovatsus:
After uninstalling .NET Core SDK 2.1.101, master now builds, and now I have the same error that shows up on CI instead of things building locally:

    Failed   Respects encoding when specified
    Error Message:
     System.InvalidOperationException : tried to skip The input sequence has an insufficient number of elements. 1 past the end of the seq
    Stack Trace:
       at Microsoft.FSharp.Collections.SeqModule.Skip@1322.GenerateNext(IEnumerable`1& next)
       at Microsoft.FSharp.Core.CompilerServices.GeneratedSequenceBase`1.MoveNextImpl()
       at Microsoft.FSharp.Core.CompilerServices.GeneratedSequenceBase`1.System-Collections-IEnumerator-MoveNext()
       at Microsoft.FSharp.Collections.SeqModule.Head[T](IEnumerable`1 source)
       at FSharp.Data.Tests.CsvProvider.Respects encoding when specified()

dotnet --version says it's 2.1.103, which is a bit odd, no idea where that came from :)

The errors on FSharp.Data.Tests.sln are still there, though

ovatsus:
Another side effect of moving to the new nuget references, is that now just opening FSharp.Data.sln and building inside VS2017 doesn't work anymore. It used to auto restore packages with paket, but now because it doesn't use it, the typeprovidersdk files are not restored automatically. Can we import the paket targets like before, or will it cause problems because the references aren't using paket anymore?

dsyme:
Just to check status

* [ ] Failing test - need to finalise what to do here
* [ ] "Master not building tests after updating VS to 15.6.4"
* [ ] Paket files not being restored on open of FSharp.Data.sln

thanks

ovatsus:
Master not building after update was fixed by uninstalling version 2.1.101 of .net core sdk.
The failing test it's the cp292 one, but it's failing on the .net sdk version, which it shouldn't
I think there's also a problem that the command line build is not putting the binaries in the same place than VS, and then when you open VS and try to build everything gets weird, and you can't run/debug any tests in the VS test explorer

dsyme:
> The failing test it's the cp292 one, but it's failing on the .net sdk version, which it shouldn't

I'll sort this out.  It's failing on net461 because the compiler now runs using .NET SDK, even when compiling for net461

> I think there's also a problem that the command line build is not putting the binaries in the same place than VS, and then when you open VS and try to build everything gets weird, and you can't run/debug any tests in the VS test explorer

OK, will add to list of things

dsyme:
> Master not building after update was fixed by uninstalling version 2.1.101 of .net core sdk.

Hmmm... I would like to know why this was needed...  I have 2.1.103 after installing VS 15.6.4

dsyme:
I added the import of Paket.Restore.targets, it seems to do the trick for the paket files

ovatsus:
I had 101 after installing 1.5.6.4, then when I uninstalled 101 manually iI magically still had 103 installed afterwards, and it works fine with that

On 28 Mar 2018 17:35, Don Syme <notifications@github.com> wrote:

Master not building after update was fixed by uninstalling version 2.1.101 of .net core sdk.

Hmmm... I would like to know why this was needed... I have 2.1.103 after installing VS 15.6.4

baronfel:
That should be able to be fixed by adding a global.json with the sdk version we want to force at build-time to the root of the repo.

so something like:
```
{
    "sdk": {
         "version": "2.1.103"
    }
}
```

baronfel:
Also @forki fixed the multiple-runtime version selection bug that @dsyme reported as part of this work yesterday. We could try to revert the PackageReference stuff now and be back to only paket to reduce confusion.

ovatsus:
Ok, now on a clean repo I open FSharp.Data.sln with VS and everything works out of the box, but after build finishes if I then open FSharp.Data.Tests.sln adn try to build it, I get all these errors, which prevent me from debugging any test

image

tests/FSharp.Data.Tests/CsvProvider.fs
```diff
@@ -306,7 +306,7 @@ let ``Uses UTF8 for sample file when encoding not specified``() =
     let row2 = utf8.Rows |> Seq.skip 1 |> Seq.head
     row2 |> should equal (2, "NaN (�񐔒l)")
 
-#if !NETCOREAPP2_0 // "No data is available for encoding 932. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method."
+#if USE_MSBUILD // only vali when running with the .NET Framework compiler
```
knocte: typo "vali" -> "valid"

ovatsus:
Are both Travis and AppVeyor only using the sdk version? Would be good to keep at least one CI config using MSBuild to make sure we don't break anything there

eriawan:
@ovatsus and @dsyme 

Just a suggestion on this problem:
> I think .NET Core just doesn't support this encoding.
> This means that when F# design-time tooling runs using .NET Core, FSharp.Data design-time tooling can't be used with files that have this encoding.

About the error on codepages when using .NET Core 2.0, I have also experienced that problems.

To fix this, I add nuget package of `System.Text.Encoding.CodePages`. 
On the code that has the codepage problem, I initialize a codepage provider using this code:

`System.Text.Encoding.RegisterProvider(System.Text.CodePagesEncodingProvider.Instance)`


Then that code will register the codepages that previously available/supported in .NET Framework 4.6.x/4.7.x.

I use that fix because System.Text.CodePagesEncodingProvider is not shipped as part of current .NET Core 2.0 SDK, it's marked as ".NET Core 2.0 + Platform extensions".
We can also check this on https://apisof,net: (https://apisof.net/catalog/System.Text.CodePagesEncodingProvider)

image

Looking at that table, System.Text.CodePagesEncodingProvider will be shipped internally (available by default) in .NET Core 2.1.

For now, on .NET Core 2.0, I suggest to use the fix above.
Please let me know if it still doesn't fix the codepage problem.

dsyme:
@eriawan Thanks for the info - the situation is complicated because this is code executing at design time (the FSharp.Data.DesignTime.dll TPDTC), and the F# compiler in the .NET SDK (which runs with .NET Core) is having trouble loading dependencies for type providers, see https://github.com/Microsoft/visualfsharp/issues/4579.  So we can't make the TPDTC dependent on other packages at right now

dsyme:
> Are both Travis and AppVeyor only using the sdk version? Would be good to keep at least one CI config using MSBuild to make sure we don't break anything there

Travis configuration is setting USE_MSBUILD when using Mono.  

AppVeyor is not but we could add a matrix

dsyme:
AppVeyor is now passing, with both .NET SDK configuration (the default) and USE_MSBUILD=1 

Travis is having an issue because the .NET SDK was not being installed, seeing if that's fixed now

dsyme:
@baronfel The problem with FSharp.Data.Tests.sln is annoying, it looks like the VS tooling doesn't like references to out-of-solution projects for new-style project files.

I'm not sure what the right solution to this is.  One would be not to have `to require a local build of the packages and use a PackageReference.  But this feels wrong.

I've sent this query to @tihan

> Hey Will,
> 
> We have an issue in FSharp.Data involving a ProjectReference to a project not in the solution, e.g.
> 
>     FSharp.Data.sln  
>                FSharp.Data.fsproj
>                FSharp.Data.DesignTime.fsproj
>
>    FSharp.Data.Tests.sln  
>                FSharp.Data.DesignTime.Tests.fsproj  contains project reference to FSharp.Data.fsproj and FSharp.Data.DesignTime.fsproj
> 
> Things build OK from the command line with both “msbuild.exe” and “dotnet build” but when we open FSharp.Data.Tests.sln (or FSharp.Data.DesignTime.Tests.fsproj) using 15.6.4 we get this:
> image
> 
>  What’s the expected behaviour when opening projects that contain references to other projects that are not in the solution?  Is there are workaround for this? We can’t easily put everything in one solution because the type provider design-time component gets locked when in use from the test projects

dsyme:
Apart from the issue of opening `FSharp.Data.Tests.sln` and getting Travis green I think we're clear of known issues now (though we should double check everything! :) )

ovatsus:
Yes, that should be it

baronfel:
I have similar IDE issues in VS Code when launching the test solution, at least until I debug build FSharp.Data.DesignTime/FSharp.Data. The error explicitly asks for debug binaries to be built, and once that's done Ionide begins working again.

dsyme:
@baronfel @baronfel I pushed a fix so that using both FSharp.Data.sln and FSharp.Data.Tests.sln work for build/edit.  I'm pretty sure it's a sound approach - we use ``Reference`` nodes and layout the files under `bin` just like the nuget package.

ovatsus:
To be clear, I don't think that's a deal breaker, as soon as CI is passing I think this should be good to go.

@baronfel , @dsyme, are any of you looking at the mono test failure?

baronfel:
The OSX, latest mono, dotnet-cli failures seem to be transient and due to invalid json, I'm not sure they represent an actual problem in the code.

The Linux, weekly mono, msbuild failures are more concerning. I can't replicate them locally on OSX/mono 5.10 as of yet.

dsyme:
These are the failures on Linux Mono weekly:
```
Error : FSharp.Data.Tests.Http.Web request's timeout is used
[37m[31mSystem.NullReferenceException : Object reference not set to an instance of an object
[37m[31m  at FSharp.Data.Tests.Http.Web request's timeout is used () [0x00026] in <5abc27ca49fdba00a7450383ca27bc5a>:0 
  at (wrapper managed-to-native) 

Failed : FSharp.Data.Tests.HttpIntegrationTests.all details of the response should be available
 Expected: <Text "Some JSON or whatever">
 But was:  <Text "">
```

dsyme:
@ovatsus @baronfel I don't think the two failures in Mono weekly should stop us integrating.  They may be known bugs in Mono that are fixed before release, or we may need to record bugs against Mono.  I'll mark the weekly as a configuration we can ignore errors on (I don't like doing that of course)

dsyme:
We now only have these failures on Mono+latest+OSX+USE_MSBUILD.

It's very odd that this configuration specifically would fail when other very similar ones are passing (e.g. Mono+latest+Linux+USE_MSBUILD or Mono+5.8.0+OSX+USE_MSBUILD).

Things like NameResolutionFailure tell me it's a configuration thing.

```
1) Error : FSharp.Data.Tests.Http.Don't throw exceptions on http error
System.Net.WebException : Error: NameResolutionFailure
2) Error : FSharp.Data.Tests.Http.Setting timeout in customizeHttpRequest overrides timeout argument
System.Net.WebException : Error: NameResolutionFailure
3) Failed : FSharp.Data.Tests.Http.testFormDataBodySize(4000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
4) Failed : FSharp.Data.Tests.Http.testFormDataBodySize(20000)
  Expected: No Exception to be thrown
5) Failed : FSharp.Data.Tests.Http.testFormDataBodySize(40000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
6) Failed : FSharp.Data.Tests.Http.testFormDataBodySize(100000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
7) Failed : FSharp.Data.Tests.Http.testFormDataBodySize(200000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
8) Failed : FSharp.Data.Tests.Http.testMultipartFormDataBodySize(4000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
9) Failed : FSharp.Data.Tests.Http.testMultipartFormDataBodySize(20000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
10) Failed : FSharp.Data.Tests.Http.testMultipartFormDataBodySize(40000)
  Expected: No Exception to be thrown
  But was:  <System.AggregateException: One or more errors occurred. ---> System.Net.WebException: Error: NameResolutionFailure
11) Failed : FSharp.Data.Tests.Http.testMultipartFormDataBodySize(100000)
  Expected: No Exception to be thrown
12) Failed : FSharp.Data.Tests.Http.testMultipartFormDataBodySize(200000)
  Expected: No Exception to be thrown
13) Error : FSharp.Data.Tests.Http.Timeout argument is used
System.NullReferenceException : Object reference not set to an instance of an object
14) Error : FSharp.Data.Tests.Http.Two custom header with different names don't throw an exception
System.Net.WebException : Error: NameResolutionFailure
15) Error : FSharp.Data.Tests.Http.Web request's timeout is used
System.NullReferenceException : Object reference not set to an instance of an object

baronfel:
I agree with that route. Ideally I'd like to wrap those tests that use the network stack in a retry mechanism, specifically because we've had so many repeated problems with them.

ovatsus:
Weird, `build release` fails with this error:

    Running build failed.
    Error:
    System.Exception: Start of process fsi.exe failed. The system cannot find the file specified
       at Fake.ProcessHelper.ExecProcessWithLambdas@103-16.Invoke(String message) in D:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 103
       at Fake.ProcessHelper.ExecProcessWithLambdas(FSharpFunc`2 configProcessStartInfoF, TimeSpan timeOut, Boolean silent, FSharpFunc`2 errorF, FSharpFunc`2 messageF) in D:\code\fake\src\app\FakeLib\ProcessHelper.fs:line 103
       at Fake.FSIHelper.executeFSIWithArgs(String workingDirectory, String script, FSharpList`1 extraFsiArgs, IEnumerable`1 env) in D:\code\fake\src\app\FakeLib\FSIHelper.fs:line 186
       at FSI_0005.Build.clo@195-33.Invoke(Unit unitVar0)
       at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in D:\code\fake\src\app\FakeLib\TargetHelper.fs:line 626

But if I manually run build.fsx code inside VS it seems to work fine

PR #1124, opened 2018-02-06
thread encodings through the load methods

### LINKED ISSUE #871: Up for grabs: Add encoding parameter to load methods of CSV/XML/CSV provider
Pretty sure the problem referenced in this Gist https://gist.github.com/mitalkakaiya/18bc73acd69cee514a1c is because `StreamReader :> TextReader` constructed at line 227 in FSharp.Data.IO is always constructed with `Encoding.UTF8`. The reader skips big blocks of rows in the input file, starting after successfully reading line 10470 in the input file. 

Does not seem to be a way for user to pass alternate character encoding.

baronfel:
Should fix #871.

We do default to UTF8 if no encoding is specified, but now there are no longer any "" parameters passed as encodingStrings to the asyncReadTextAtRuntime function.

### DIFF
--- a/docs/content/library/HtmlProvider.fsx
+++ b/docs/content/library/HtmlProvider.fsx
@@ -122,11 +122,11 @@ Chart.Bar stats
 
 (*** define-output:doctorWhoChart ***)
 let [<Literal>] DrWho = 
-  "http://en.wikipedia.org/wiki/List_of_Doctor_Who_serials"
+  "https://en.wikipedia.org/wiki/List_of_Doctor_Who_episodes_(1963%E2%80%931989)"
 
 let doctorWho = new HtmlProvider<DrWho>()
 
-// Get the average number of viewers for each doctor's series run
+// Get the average number of vi ewers for each doctor's series run
 let viewersByDoctor = 
   doctorWho.Tables.``Series overview``.Rows 
   |> Seq.groupBy (fun season -> season.Doctor)
--- a/docs/content/library/HtmlProvider.fsx
+++ b/docs/content/library/HtmlProvider.fsx
@@ -126,7 +126,7 @@ let [<Literal>] DrWho =
 
 let doctorWho = new HtmlProvider<DrWho>()
 
-// Get the average number of vi ewers for each doctor's series run
+// Get the average number of viewers for each doctor's series run
 let viewersByDoctor = 
   doctorWho.Tables.``Series overview``.Rows 
   |> Seq.groupBy (fun season -> season.Doctor)
--- a/src/Csv/CsvFile.fs
+++ b/src/Csv/CsvFile.fs
@@ -93,10 +93,9 @@ and CsvFile private (readerFunc:Func<TextReader>, [<Optional>] ?separators, [<Op
 
   /// Loads CSV from the specified uri
   static member Load(uri:string, [<Optional>] ?separators, [<Optional>] ?quote, [<Optional>] ?hasHeaders, [<Optional>] ?ignoreErrors, [<Optional>] ?skipRows, [<Optional>] ?encoding) = 
-    CsvFile.AsyncLoad(uri, ?separators=separators, ?quote=quote, ?hasHeaders=hasHeaders, ?ignoreErrors=ignoreErrors, ?skipRows=skipRows) |> Async.RunSynchronously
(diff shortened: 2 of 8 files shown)


ovatsus:
Ci is failing with `The remote name could not be resolved: 'deelay.me'`. It doesn't seem related to your change, but can you have a look?

baronfel:
Will do. I saw that locally as well but pushed it up to see if it was my local DNS or something. Let's see what I can figure out :)

baronfel:
Ok, I _think_ we're good to go now. The first commit fixed the initial set of test failures which were due to `deelay.me` being unreliable, and the second commit fixed some documentation generation errors. One was due to the Doctor Who wikipedia page suddenly changing into a disambiguation page, so I chose one of the doctor who series summary links from that disambiguation page. The second was due to an errant tab whitespace character that somehow got introduced.

docs/content/library/HtmlProvider.fsx
```diff
@@ -122,11 +122,11 @@ Chart.Bar stats
 
 (*** define-output:doctorWhoChart ***)
 let [<Literal>] DrWho = 
-  "http://en.wikipedia.org/wiki/List_of_Doctor_Who_serials"
+  "https://en.wikipedia.org/wiki/List_of_Doctor_Who_episodes_(1963%E2%80%931989)"
 
 let doctorWho = new HtmlProvider<DrWho>()
 
-// Get the average number of viewers for each doctor's series run
+// Get the average number of vi ewers for each doctor's series run
```
ovatsus: NIT: you introduced a space here

src/Csv/CsvFile.fs
```diff
@@ -91,25 +92,22 @@ and CsvFile private (readerFunc:Func<TextReader>, [<Optional>] ?separators, [<Op
...
   /// Loads CSV from the specified uri
-  static member Load(uri:string, [<Optional>] ?separators, [<Optional>] ?quote, [<Optional>] ?hasHeaders, [<Optional>] ?ignoreErrors, [<Optional>] ?skipRows) = 
-    let separators = defaultArg separators ""    
-    let separators = 
-        if String.IsNullOrEmpty separators && uri.EndsWith(".tsv" , StringComparison.OrdinalIgnoreCase) 
-        then "\t" else separators
-    let readerFunc = Func<_>(fun () -> asyncReadTextAtRuntime false "" "" "CSV" "" uri |> Async.RunSynchronously)
-    new CsvFile(readerFunc, separators, ?quote=quote, ?hasHeaders=hasHeaders, ?ignoreErrors=ignoreErrors, ?skipRows=skipRows)
+  static member Load(uri:string, [<Optional>] ?separators, [<Optional>] ?quote, [<Optional>] ?hasHeaders, [<Optional>] ?ignoreErrors, [<Optional>] ?skipRows, [<Optional>] ?encoding) = 
+    CsvFile.AsyncLoad(uri, ?separators=separators, ?quote=quote, ?hasHeaders=hasHeaders, ?ignoreErrors=ignoreErrors, ?skipRows=skipRows) |> Async.RunSynchronously
```
ovatsus: You forgot to propagate the encoding parameter here

ovatsus:
Sorry, meant to review this earlier but it slipped my mind. Left a couple of comments, you have a small bug.
Can you also add an entry in ReleaseNotes? (Create a new 2.4.5 section with "Unreleased" as the date)

src/Csv/CsvFile.fs
```diff
@@ -91,25 +92,22 @@ and CsvFile private (readerFunc:Func<TextReader>, [<Optional>] ?separators, [<Op
...
 
   /// Loads CSV from the specified uri asynchronously
-  static member AsyncLoad(uri:string, [<Optional>] ?separators, [<Optional>] ?quote, [<Optional>] ?hasHeaders, [<Optional>] ?ignoreErrors, [<Optional>] ?skipRows) = async {
-    let separators = defaultArg separators ""    
+  /// This doesn't actually load the file async, so why is it here? The `Async.RunSynchronously` ruins everything.
```
ovatsus: It only forces to download the entire file when you iterate .Rows to to get the first item, so in that sense it's still async. To be truly async you would have to use something like AsyncSeq which would complicate the usage quite a bit for the common case. I'd be welcome to have one variation that uses it though, but it's actually quite complicated to have both (I tried long time ago), and then you add a dependency to another library.
Can you remove the comment in the next iteration? It's using triple slash so it's going to the docs

baronfel:
Thanks for the review! I'll address those points today.

ovatsus:
Thanks

PR #1134, opened 2018-02-28
Send JSON requests with content type charset set explicitly to UTF-8

Regent:
Proposed fix.
Need an opinion on `ContentTypeWithCharset` vs `ContentTypeWithEncoding` - personally I prefer to have the latter so you can't accidentally provide charset string instead of content type string and vice versa, but debatable.


ovatsus:
I would keep just ContentTypeWithEncoding

ovatsus:
Can you update releasenotes.md?

PR #1135, opened 2018-02-28
Parsing JSON arrays when culture is using comma as a decimal separator

Regent:
Personally I would have stricter implementation, maybe even up to a letter of [json.org](https://www.json.org/) specification...


ovatsus:
@Regent , would you mind changing to the stricter implementation? Ever allowing commas in json arrays was a weird thing that was done for apiary, and I think it was a mistake. Since the next release will have a bump in the major version number, I think we should fix this fully and take the breaking change.

ovatsus:
BTW, this behaviour of looking at the current culture for the separator was introduced in 2013 (https://github.com/fsharp/FSharp.Data/commit/7bc8693a3f3c626718050a2da893ac36975594e8#diff-6b054d04c7baa359c201d6389821678a), I probably didn't know what I was doing at the time :)

Regent:
Sure, I'll do that.

ovatsus:
Ah, already did it and merged :)

Regent:
No worries, I'll create new PR before the release :)

ovatsus:
What I meant was, I did those changes myself and pushed to your branch before merging, so it's all done

Regent:
Oh, I see. Thanks! :)

PR #1136, opened 2018-03-09
implement prefer header per RFC 7240

### LINKED ISSUE #1133: Add Prefer header to HttpRequestHeaders

GHvW:
Fixes #1133.
Do you have any suggestions about the wording of the documentation?


baronfel:
There's a bit more work on this that @GHvW pointed out to me after rereading the spec around the key-value pairs. 

From Slack:
> Garrett: 
> I'm not going to lie, their examples are terrible imo. Too much left to be inferred, maybe just to the uninitiated though - like me. Check out the bottom of page 6 and beginning page 7 examples. Ex 1 has by my interpretation 3 prefer preferences, 2 on the same header one on a separate one. Example 3 is one preference with one parameter for that preference by my interpretation. Thoughts?
> chethusk:
> ohhhhh I see
> the key-value preferences are actually a key-set-of-values preference
> where the set-of-values are semicolon-delimited (edited)
> something more like:
> ```type Preference =
> | Flag of flag: string
> | KeyValues of key: string * values: string seq
>
> let Prefer (preferences: Preference seq) = ......
 > ```
> so then our toString for the `Preference` type changes a bit to deal with the fact that `values` is a seq and not a single string:
> 
> ```type Preference =
> | Flag of flag: string
> | KeyValues of key: string * values: string seq
> with
>  override this.ToString() = 
>    match this with
>    | Flag flag -> flag
>    | KeyValues(key, values) -> 
>      let valuesAsStr = values |> String.concat ";"
>      sprintf "%s: %s" key valuesAsStr```
>```

ovatsus:
I think we can do this without introducing a new type, and if I interpreted the RFC correctly, when there's a value it needs to be inside double quotes. This is what I was thinking:

    let Prefer (prefer: (string * (#seq<string> option)) seq) =
        "Prefer", 
        prefer 
        |> Seq.map (function
            | key, None -> key
            | key, Some values -> sprintf "%s=\"%s\"" key (String.concat "," values))
        |> String.concat "; "

example usage

    Prefer ["behavior1", None; "behavior2", Some ["A"]; "behavior3", Some [ "B"; "C"]]
    // returns Prefer: behavior1; behavior2="A"; behavior3="B,C"

(I filed this issue when I found myself needing something like behavior3)

ovatsus:
Or maybe this?

    let Prefer (prefer: (string * #seq<string>) seq) =
        "Prefer", 
        prefer 
        |> Seq.map (fun (key, values) ->
            if Seq.isEmpty values
            then key
            else sprintf "%s=\"%s\"" key (String.concat "," values))
        |> String.concat "; "

GHvW:
This is how I interpreted the RFC (I definitely don't advocate doing it this way, it's really ugly, but I think it's a decent representation in code of how I interpreted the RFC)
```
type Token = Token of name:string * value:string option
with
    override this.ToString() =
        match this with
        | Token (name, None) -> name
        | Token (name, Some value) -> sprintf "%s=%s" name value

let Prefer (preferences: (Token * #seq<Token> option) seq) =
    "Prefer",
    preferences
    |> Seq.map (fun preference ->
        match preference with
        | pref, None -> pref.ToString()
        | pref, Some parameters -> sprintf "%s; %s" (pref.ToString()) (parameters |> Seq.map (fun parameter -> parameter.ToString()) |> String.concat "; "))
    |> String.concat ", " 
```
```
Prefer [Token("behavior1", None), None; Token("behavior2", Some "value"), Some [Token("A", Some "X"); Token("B", Some "Y")]; Token("behavior3", Some "\"weird value\""), None]
//returns Prefer: behavior1, behavior2=value; A=X; B=Y, behavior3="weird value"
```
This Prefer header has 3 preferences
behavior1 has no value associated with it and no parameters.
behavior2 have a value associated with it, and has 2 parameters, each with a corresponding value
behavior3 has a quoted-string value, but no parameters.

where:
-Every Prefer header can have one or more preferences, each preference separated by a comma ","
-Every preference can have 0 or multiple parameters, each parameter separated from it's preference and other parameters by a semi-colon ";"
-Every preference or parameter is a token with a name and an optional value, if it has a value, the name and value are separated by an equals "=". A value can optionally be enclosed in quotes

I got the part about values being in quotes or not from the bottom of page 4 of the RFC:
"  For both preference token names and parameter names, comparison is
   case insensitive while values are case sensitive regardless of
   whether token or quoted-string values are used."


It might be easier and cleaner to make all preferences and parameters a single string provided by the user. If they want "handling=lenient" they would provide that full string instead of "handling", "lenient".
Ex:
```
let Prefer (preferences: (string * #seq<string> option) seq) =
    "Prefer", 
    preferences 
    |> Seq.map (fun preference ->
        match preference with
        | pref, None -> pref
        | pref, Some parameters -> sprintf "%s; %s" (pref) (parameters |> String.concat "; ")) 
    |> String.concat ", "
```
```
Prefer ["handling=lenient", None; "response-async", None; "return=minimal", Some ["foo"; "bar=value"; "baz=blah"]]
//returns Prefer: "handling=lenient, response-async, return=minimal; foo; bar=value; baz=blah"
```
Sorry for the wall

Thoughts? What would you prefer?

ovatsus:
I read the RFC and you're right, I got the `,` and `;` wrong, your implementation is almost the correct one. You just need `sprintf "%s=\"%s\"" name value` instead of `sprintf "%s=%s" name value` in the `| Token (name, Some value)` case

However, the header that I wanted to use that led me to submit this issue was `Prefer: exchange.behavior="extension1,extension2"`

On your first version, that would force me to use `Prefer [Token("echange.behavior", Some "extension1,extension2"), None]`, on the second one `Prefer ["echange.behavior=\"ext1,ext2\"", None]`, none seems like an improvement over keeping it simple and just doing `let Prefer (prefer:string) = "Prefer", prefer` :(

PR #1165, opened 2018-05-17
Added application/json-rpc to mimetype and isText.

deapsquatter:
The application/json-rpc should be treated as text. See http://www.jsonrpc.org/historical/json-rpc-over-http.html.


ovatsus:
Thanks

Issue #1163, opened 2018-05-11
Question: Is there any reason there is no module with standard HTTP status codes?


NatElkins:
When using the HTTP utilities, I noticed that there were constants for common headers, but not status codes.  Would that be a welcome PR?  I would create the status code constants as per https://en.wikipedia.org/wiki/List_of_HTTP_status_codes .

ovatsus:
Yes, that would be a nice addition

0x456464:
I'll take a look at this now.

0x456464:
Submitted a pull request.

ovatsus:
Seems you both sent PRs to do the same

PR #1168, opened 2018-05-27
Add http status codes

NatElkins:
Fixing https://github.com/fsharp/FSharp.Data/issues/1163.

I used the following script to generate most of the code:

```fsharp
#r "../../packages/FSharp.Data/lib/net45/FSharp.Data.dll"
#r "../../packages/FSharp.Data/lib/net45/FSharp.Data.DesignTime.dll"

open System
open System.Globalization
open FSharp.Data
open System.Text.RegularExpressions
open System.Text

let sc = new HtmlProvider<"https://en.wikipedia.org/wiki/List_of_HTTP_status_codes">()

/// Returns a string that represents the code
/// for a literal property.
let printLiteralCode (code:string) (name:string)  =
    let filterPunctuation = String.filter (Char.IsPunctuation >> not)
    let propName = 
        name.Split ' '
        |> Array.map (filterPunctuation >> CultureInfo.CurrentCulture.TextInfo.ToTitleCase)
        |> fun words -> String.Join("",words)
    sprintf
        """/// %s - %s
let [<Literal>] %s = %s""" name code propName code

let sb = StringBuilder()
let pattern = """([0-9]{3})\s([a-zA-Z\s'\-]+)\b"""

sc.Html.Descendants("dt")
|> Seq.iter(fun node ->
    let innerText = node.InnerText().Trim()
    let m = Regex.Match(innerText,pattern)
    if m.Groups.Count = 3 then
        let code = m.Groups.[1].Value
        let name = m.Groups.[2].Value
        ignore (sb.AppendLine (printLiteralCode code name))
)

System.IO.File.WriteAllText("/path/to/StatusCodes.fs", sb.ToString())
```

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -272,6 +272,134 @@ type HttpResponseWithStream =
     Headers : Map<string,string>
     Cookies : Map<string,string> }
 
+/// Constants for HTTP status codes.
+/// Based on https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
+module HttpStatusCodes =
+    /// Continue - 100
+    let [<Literal>] Continue = 100
+    /// Switching Protocols - 101
+    let [<Literal>] SwitchingProtocols = 101
+    /// Processing - 102
+    let [<Literal>] Processing = 102
+    /// Early Hints - 103
+    let [<Literal>] EarlyHints = 103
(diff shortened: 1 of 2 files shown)


ovatsus:
Thanks

PR #1171, opened 2018-06-16
Status codes added to Http.fs

0x456464:
I've added a module of HTTP Status Codes to Http.fs for issue #1163. I have only included official status codes, and did not include the chocolate teapot (418) Easter egg code (I did consider it!). 

The comments are abbreviated or complete copies of the status code descriptions found here -  https://en.wikipedia.org/wiki/List_of_HTTP_status_codes. 

Let me know if you would like anything changed, or if I missed anything - this is my first pull request!


ovatsus:
I merged @NatElkins implementation, keeping his names, but then also merged yours as your comments were nicer

Issue #1167, opened 2018-05-26
Crashes when loading an html page with a table with a zero length name


tplatt:
In NameUtils.fs the current code is:

let uniqueGenerator niceName =
  let set = new HashSet<_>()
  fun name ->
    let mutable name = niceName name
    while set.Contains name do 

The following change fixes the issue

let uniqueGenerator niceName =
  let set = new HashSet<_>()
  fun name ->
    let mutable name:String = niceName name
    if name.Length = 0 then name <- "Unnamed"
    while set.Contains name do

0x456464:
Pull request - https://github.com/fsharp/FSharp.Data/pull/1173

PR #1173, opened 2018-06-17
Issue #1167

0x456464:
Issue: Crashes when loading an html page with a table with a zero length name 

Fix: If uniqueGenerator is passed a name value of length 0 the name is defaulted to 'Unnamed'.


ovatsus:
Can you add a UT with a HTML file that would repro #1167?

ovatsus:
Ping on the UT :)

PR #1178, opened 2018-06-30
Fix handling of excaped charaters in string literals within script tag

serhiiz:
When parser is in a script mode and encounters " or ' it detects string literal and will consume characters until it finds matching " or ' correspondingly. However, " or ' can be escaped, so that they loose their original meaning and should be ignored by the parser at that point. Parser should look for next " or '.


ovatsus:
Thanks

Issue #1040, opened 2017-04-06
Http.Request fails with "part of the cookie is invalid" on VSTS API


kberridge:
When I query the Visual Studio Team Services (VSTS) API:

`Http.RequestString("https://organization.visualstudio.com/DefaultCollection/Project/_apis/wit/queries", headers = [ "Authorization", "Basic base64encodeduseraccesstoken"])`

I get the following error:

```
Unhandled Exception: System.Net.CookieException: The 'Value'='{"PersistentSessionId":"3cd7392c-35f2-4fac-a31a-533d33d6ca6f","PendingAuthenticationSessionId":"0
0000000-0000-0000-0000-000000000000","CurrentAuthenticationSessionId":"00000000-0000-0000-0000-000000000000"}' part of the cookie is invalid.
   at System.Net.Cookie.VerifySetDefaults(CookieVariant variant, Uri uri, Boolean isLocalDomain, String localDomain, Boolean set_default, Boolean isThrow)
   at System.Net.CookieContainer.Add(Uri uri, Cookie cookie)
   at <StartupCode$FSharp-Data>.$Http.InnerRequest@930-5.Invoke(WebResponse _arg2) in C:\Git\FSharp.Data\src\Net\Http.fs:line 938
   at Microsoft.FSharp.Control.AsyncBuilderImpl.args@825-1.Invoke(a a)
```

I switched to WebClient.DownloadString() and that is working.  I also tried it in Postman hoping I could attach some more information about what the server is returning but Postman says the server didn't return a cookie.

I also found this possibly related open issue #982, but that error was "The 'Name'='' part of the cookie is invalid." which seemed different enough so I thought opening a new issue was the right move.

ovatsus:
Hi @kberridge, do you want to try to put a breakpoint and see what is failing and try to fix it?

ovatsus:
I did a similar request to my organisation VSTS and couldn't repro the bug. If you can still repro it, please provide the cookie value that breaks and reopen this issue

ovatsus:
@Peter360, can you check if it's working with 3.0? There was a cookie related fix

Peter360:
Yes the problem persists.

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Program.fs" />
[...]

PR #1183, opened 2018-07-19
Avoid errors with cookie value not supported by RFC6265

devcrafting:
Fix partially #1040.
They are two ways to avoid the issue :

* Do not use CookieContainer in parameter (an internal one is used, but it avoids the issue)
* Or, if you don't care about the cookie that causes the issue, you can set optional silentCookieError parameter

Still if you need this malformed cookie (should be URL encoded by the server following RFC6265) and still want to have the facility to pass cookieContainer through several request to avoid manuel cookie management, then I don't have a solution.

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1489,12 +1489,13 @@ type Http private() =
                 [<Optional>] ?body,
                 [<Optional>] ?cookies,
                 [<Optional>] ?cookieContainer,
+                [<Optional>] ?silentCookieErrors,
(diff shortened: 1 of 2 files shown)


devcrafting:
Weird error on AppVeyor, seems unrelated to code. Could you run it again ?

devcrafting:
Would anyone have chance to have a look at this PR ? (I know it is holidays period ^^)
Let me know. I use it with a private Nuget repo, sounds good :).
Thanks.

devcrafting:
No worries and thanks for its integration ;).

PR #1185, opened 2018-07-24
Don't silently convert floats to integers in JsonProvider.

EduardOstrovsky:
Sufficiently large float values will end up converted as Int.MinValue and succeed


ovatsus:
Thanks, I've merged this, will be part of 3.0.0-beta5

ovatsus:
Had to do a small tweak to the UT so it passed on my machine ("+39" vs "+039") - https://github.com/fsharp/FSharp.Data/commit/7ca8802f26e7b4123fe18b509359a78ab8dc0265

Issue #1179, opened 2018-07-10
Large JSON hangs FSI


jdh30:
Due to the `[<StructuredFormatDisplay("{_Print}")>]` in the definition of `JsonValue` F# Interactive prints JSON in its entirety. Even for moderately large datasets this can take a very long time. For example, I am dealing with a 41MB JSON file and F# Interactive hangs for several minutes printing it every time I evaluate an expression that results in JSON.

ovatsus:
Feel free to send a PR changing the _Print method to put a cap on the amount of data and printing a ... instead

zakaluka:
Submitted initial PR to cap value printouts at 512 characters (printout will be 509 characters + 3 dots)

zakaluka:
If we don't want to print anything to the console for these objects, that makes complete sense.  @ovatsus which way would you like to proceed?

PR #1186, opened 2018-07-28
Put cap on printing large values #1179

zakaluka:


zakaluka:
The AppVeyor failure seems to be caused due to access restriction to `dotnet.exe`.  https://ci.appveyor.com/project/ovatsus/fsharp-data/build/3.0.0-beta4-a1807280821/job/2lol2nl6w81un6fe#L2628

giacomociti:
I'm afraid the `ToString` method of `XmlElement` is affected since it relies on `_Print`. For json and html I see no problem because apparently it's the other way around: `_Print` calls `ToString`.

zakaluka:
I see what you mean.  I've updated is accordingly (so that `x.ToString()` returns the entire string, and `_Print` relies on `x.ToString()`).

ovatsus:
#1179

ovatsus:
Why did you remove StructuredFormatDisplay? If you do that then FSI will display the internal structure, instead of the nice pretty printed output. I think what we should do instead, is have _Print calling .ToString() and then capping the returned string at 512 characters

zakaluka:
Undid my previous changes, and capped the output.

src/Xml/XmlRuntime.fs
```diff
@@ -57,7 +57,6 @@ open System.Xml.Linq
 #nowarn "10001"
 
 /// Underlying representation of types generated by XmlProvider
-[<StructuredFormatDisplay("{_Print}")>]
```
ovatsus: Revert this line

src/Xml/XmlRuntime.fs
```diff
@@ -67,12 +66,12 @@ type XmlElement =
   /// [omit]
   [<EditorBrowsableAttribute(EditorBrowsableState.Never)>]
   [<CompilerMessageAttribute("This method is intended for use in generated code only.", 10001, IsHidden=true, IsError=false)>]
-  member x._Print = x.XElement.ToString()
+  member x._Print = x.ToString()
```
ovatsus: Do the same capping here

src/Xml/XmlRuntime.fs
zakaluka: I've reverted the removal of that line.  Do you want me to remove the line itself?

Issue #1181, opened 2018-07-16
XSD date handling


giacomociti:
I'd like to discuss [this issue](https://github.com/fsprojects/FSharp.Data.Xsd/issues/24) here, now that XSD support is merged. I feel uneasy ignoring it; we know that type providers are more geared towards reading than writing, but providing a ctor with a `System.DateTime` parameter that leads to creating invalid xml is simply wrong.

I see two alternatives: the easy one is to drop support for `xs:date`, falling back to `string`; another option may be to rely on #980 (once it gets merged), mapping `xs:dateTime` to `DateTimeOffset` and `xs:date` to `DateTime` and, [when formatting xml dates](https://github.com/fsharp/FSharp.Data/pull/980/files#diff-0c8d16397cad995a7e34291cefb11b22R211), use the proper format (including time for `DateTimeOffset`  and excluding it for `DateTime`). This is a bit hacky, but would solve the issue with no impact on the other (non-XML) type providers.

ovatsus:
I think relying on #980 would make sense. I want to merge that one and #1027 next

ovatsus:
@giacomociti , #980 is merged, can you update the XSD support to take advantage of it?

PR #1197, opened 2018-09-03
Improve XSD date handling (#1181)

giacomociti:
distinguish date and dateTime according to #1181

### DIFF
--- a/src/Xml/XmlRuntime.fs
+++ b/src/Xml/XmlRuntime.fs
@@ -1,4 +1,4 @@
-﻿// --------------------------------------------------------------------------------------
+// --------------------------------------------------------------------------------------
 // XML type provider - methods & types used by the generated erased code
 // --------------------------------------------------------------------------------------
 namespace FSharp.Data
@@ -208,7 +208,7 @@ type XmlRuntime =
             | _ ->
                 match v with
                 | :? string         as v -> v
-                | :? DateTime       as v -> v.ToString("O", cultureInfo)
+                | :? DateTime       as v -> v.ToString("yyyy-MM-dd")
                 | :? DateTimeOffset as v -> v.ToString("O", cultureInfo)
                 | :? int            as v -> strWithCulture v
                 | :? int64          as v -> strWithCulture v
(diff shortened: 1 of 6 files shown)


ovatsus:
Yes, command line build and CI is currently broken, though building and running tests in VS still works

src/Xml/XmlRuntime.fs
```diff
@@ -208,7 +208,7 @@ type XmlRuntime =
             | _ ->
                 match v with
                 | :? string         as v -> v
-                | :? DateTime       as v -> v.ToString("O", cultureInfo)
+                | :? DateTime       as v -> v.ToString("yyyy-MM-dd")
```
ovatsus: Why this change?

ovatsus:
Can you update one of the xsd test files used in SignatureTestCases.config and regenerate the expected files so we have coverage for this? (you can regenerate by evaluating SignatureTests.fs in FSI and running generateAllExpected()

src/Xml/XmlRuntime.fs
giacomociti: If we make `System.DateTime` correspond to `xs:date` (and `System.DateTimeOffset` correspond to `xs:dateTime`) when formatting a `xs:date` we have to avoid the time component in order to produce valid xml according to the XML Schema specification. On the other hand, when formatting a `xs:dateTime` it's ok to use the `"O"` format that is compliant with the W3C specs for `xs:dateTime`

giacomociti: The problem is, even if we emit zeros for the time component, the resulting xml would be invalid.

Maybe we could use the "yyyy-MM-dd" format when the time component is zero and use the "O" format otherwise.

ovatsus:
Would you mind also updating the contributing.md adding that note about generateAllExpected?

src/Xml/XmlRuntime.fs
ovatsus: That would work

ovatsus:
Looks good, can you update release notes and add a test that asserts the toString output doesn't have the time component?

ovatsus:
Thanks

Issue #1195, opened 2018-09-01
Current master doesn't build


ovatsus:
My guess the latest VS2017 update broke it, now when I try to do `build All` I get a bunch of errors like this:

`JsonExtensionsTests.cs(46,32): error CS1705: Assembly 'FSharp.Data' with identity 'FSharp.Data, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' uses 'FSharp.Core, Version=4.5.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'FSharp.Core' with identity 'FSharp.Core, Version=4.4.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' [C:\Git\FSharp.Data\tests\FSharp.Data.Tests.CSharp\FSharp.Data.Tests.CSharp.csproj]`

ovatsus:
It seems FSharp.Data is being compiled against 4.5.0.0, instead of against 4.3.1.0 as it's supposed to, and the version is also being set as 1.0.0.0, instead of 3.0.0.0

@dsyme, @baronfel, help please! :)

baronfel:
I hope to have some time to look at this later this evening. Maybe it's something simple :D

ovatsus:
I can build both solutions in VS2018 and run all the tests in the VS test runner and it all works, it's just the command line builds which are broken

baronfel:
Few notes:

failures are only on the net45 framework, and some go away when I refactor the packagereferences to include a System.Runtime reference for the net45 target:
```
<ItemGroup Condition="'$(TargetFramework)' == 'net45'" >
    <PackageReference Include="FSharp.Core" Version="4.0.0.1" />
    <PackageReference Include="System.Runtime" Version="4.0.*" />
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
    <PackageReference Include="FSharp.Core" Version="4.3.4" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.8.0" PrivateAssets="All" />
    <PackageReference Include="SourceLink.Embed.PaketFiles" Version="2.8.0" PrivateAssets="All" />
    <DotNetCliToolReference Include="dotnet-sourcelink" Version="2.8.0" />
  </ItemGroup>
```

[...]

PR #1200, opened 2018-09-06
fix build - add DisableImplicitFSharpCoreReference

dsyme:
I don't know what has changed but it seems that after upgrading to the .NET SDK 2.1.401 we now have to use

    <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>

to disable the implicit FSharp.Core reference.  Previously this was disabled automatically if you have an FSharp.Core reference in your project file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ You can see the version history [here](RELEASE_NOTES.md).
 
 ## Building
 
-- Install .NET SDK 2.1.100 or higher
+- Install .NET SDK 2.1.401 or higher
 - Build FSharp.Data.sln and FSharp.Data.Tests.sln in Visual Studio 2017, Visual Studio 2017 for Mac (previously Xamarin Studio), or MonoDevelop. You can also use the FAKE script:
 
   * Windows: Run *build.cmd* 
--- a/build.fsx
+++ b/build.fsx
@@ -33,7 +33,7 @@ let gitOwner = "fsharp"
 let gitHome = "https://github.com/" + gitOwner
 let gitName = "FSharp.Data"
 
-let desiredSdkVersion = "2.1.100"
+let desiredSdkVersion = "2.1.401"
 let mutable sdkPath = None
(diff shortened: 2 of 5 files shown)


dsyme:
Hopefully fixes https://github.com/fsharp/FSharp.Data/issues/1195

dsyme:
Add visualfsharp issue here: https://github.com/Microsoft/visualfsharp/issues/5595

ovatsus:
I also noticed that we have these 3 lines on all fsprojs

    <PackageReference Include="FSharp.Core" Version="4.0.0.1" Condition="'$(TargetFramework)' == 'net45'" />
    <PackageReference Include="FSharp.Core" Version="4.2.3" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
    <PackageReference Include="FSharp.Core" Version="4.3.4" Condition="'$(TargetFramework)' == 'netstandard2.0'" />

The last one should be deleted, right?

ovatsus:
Travis is still failing, but this fixes the main issue, so merging this

dsyme:
> I also noticed that we have these 3 lines on all fsprojs

Eventually we should just use FSharp.Core 4.3.4 everywhere, though it depends if you still want the TP usable in VS2015 (which I suppose we do)

dsyme:
> Actually, since now fsharp.core is coming through a nuget package, wouldn't it still work on all old versions of VS even if we referenced 4.5? (Assuming newer fsharp.core versions dont depend on newer .net version BCLs)

TPDTCs have to load into tooling which use whatever version of FSharp.Core they use.  so a TPDTC should depend on an early FSharp.Core that is lower than anyof the FSHarp.Core used by tooling it will load into.

dsyme:
> The last one should be deleted, right?


Yes, now I think about it the last one should be deleted, the TPDTC should depend on early FSharp.Core as mentioned above

Issue #1091, opened 2017-10-18
HtmlDocument cannot parse document with some JavaScript RegEx replace code


eugbaranov:
Having regex replace in JavaScript for left angle bracket results in the parser not being able to parse HTML.
Consider the following HTML:
```html
<head>
    <script>
        "".replace(/</g, "")
    </script>
</head>
<body>
</body>
```
The following will throw an exception:
```fsharp
let test = html |> HtmlDocument.Parse
test.Body()
```

ovatsus:
Do you want to try and send a PR to fix it?

Regent:
I'll give it a try.

colinbull:
10 months on, but I think I have a fix here. @eugbaranov can you check? The original failing test now passes, and only one of the other signature files required a regen.

eugbaranov:
`HtmlParser` changes looks good to me.

PR #1204, opened 2018-09-09
Glchapman master

colinbull:
Hey, a (potential) fix for issue #1091 (Javascript regex) and the extra case you added #1204. I have pushed it to you, so it stays as your PR. Sorry, it took so long. I started a fix then completely forgot to actually do anything about it.


colinbull:
@ovatsus, @dsyme - Note the changes in the `*.Tests` files. I have removed the net461 as a target framework. My build was failing if `net461` was included and I'm pretty sure those frameworks are mutually exclusive anyway. Was there a particular reason for this?

baronfel:
The big difference there is that we won't run the tests against a full-framework target now. It's generally good to have multiple test targets even if your core library is `netstandard2.0` just to make sure that you test on all platforms you intend to support.  I think it's worth looking us into those failures and doing what is necessary to reenable that framework.

colinbull:
Good point. Maybe I'm missing something then, my build was failing
completely, not just the tests. I was getting net461 target framework not
found or something along those lines. If I remove netcoreapp2.1 it works,
if I remove net461 it works. I didin't try but I suspect if I replaced the
net461 with net471 then it would also work.

I can obviously enable one target framework or the other, just not both in
a single project. Maybe I should just create a duplicate net461 project and
run those tests as well. Or just say that net471 is close enough(TM) to
net461 and run those tests.

Or ofcourse I'm being completely dense and have missed something super
simple.

ovatsus:
Wasn't this already fixed by #1093?

colinbull:
Nearly, there was one additional test that was failing.

ovatsus:
Can you revert the fsproj and csproj changes so I can merge just the fix, and also update release notes?

ovatsus:
Thanks

PR #1208, opened 2018-09-10
Enable empty string to be set as missingValues in inferCellType #1192

### LINKED ISSUE #1192: InferColumnTypes cannot set empty string as missingValues
**Repro steps**
sample.csv, c3 column is all empty
```
row,c1,c2,c3
1,,5,
2,4,6,
```

The following lines tried to infer the types of columns. I set missingValues parameter as `[|""|]`
```fsharp
[<Literal>]
let sample = "C:/FSharp/sample.csv"
let r = CsvFile.Load(sample)
let rTypes = r.InferColumnTypes(2,[|""|],System.Globalization.CultureInfo.GetCultureInfo(""), null, false, false)
rTypes.[3].InferedType.Name
```

**Expected outcome**
```fsharp
val it : string = "Double"
```

**Actual outcome**
```fsharp
val it : string = "String"
```

**Suggestion**
By digging through code, it is by design as empty string cannot be set as missing value. Empty string will always be treated as string. Can we loose this constraint?
https://github.com/fsharp/FSharp.Data/blob/master/src/Csv/CsvInference.fs#L130

In my use cases, csv files are quite dynamic as I don't have a fixed set of column names and column numbers. Hence schema setting won't work for me.

I'm using Deedle's Frame.ReadCsv to load csv file into data frame. But Deedle's implementation referenced CsvInference directly. I cannot set empty string to be inferred as Double.NaN.

zyzhu:
Attempt to fix #1192 
Basically I changed the order of the if statement. It shall prioritize checking values set in `missingValues` even if they contains empty string.
If empty string is not set in `missingValues`, it will still fall back to check  `String.IsNullOrWhiteSpace`


ovatsus:
I need to look a bit more into this, because we were already setting it to InteredType.Null, and that should maybe be convertible to float. Can you push a unit test that reproes this fix then I'll see if I can fix it in a different way?

zyzhu:
@ovatsus I've added a test case.

ovatsus:
Thanks. I'll have a look to see if I can fix this by changing the merging of inferred types

dsyme:
Close/open to re-run CI

Issue #1189, opened 2018-08-07
3.0.0-beta4: XmlProvider ignores attribute default in schema


unixwitch:
hello,

the new XML Schema support in 3.0.0 is really nice, thanks for integrating that.  however, i think i've found a small bug: when the schema specifies a default attribute value, the provider ignores it when parsing a document; i.e., if the attribute is not present in the element, it returns None instead of the default value.

minimal test case:

```fsharp
open FSharp.Data

type DefaultTest = XmlProvider<Schema="""
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
    <xs:element name="test">
        <xs:complexType>
            <xs:attribute name="myattr" use="optional" type="xs:string" default="a default value"/>
        </xs:complexType>
    </xs:element>
</xs:schema>""">

[<EntryPoint>]
let main argv =
    printfn "%A" (DefaultTest.Parse("""<test myattr="a value"/>""").Myattr)
    printfn "%A" (DefaultTest.Parse("""<test />""").Myattr)
    0
```

actual output:

```
Some "a value"
<null>
```

expected output:

```
Some "a value"
Some "a default value"
```

(tested with FSharp.Data 3.0.0-beta4, FSharp.Core 4.5.2, .NET Framework 4.7.1 in VS2017.)

giacomociti:
Interesting issue. Honestly I was not even aware of default values, and investigating this I learned not only about them, but also that `XDocument` has a `Validate` extension method that can do the job for us  (thanks to the `addSchemaInfo` flag):

    open System.Xml
    open System.Xml.Schema

    let parseSchema xsdText =    
        let schemaSet = XmlSchemaSet() 
        use reader = XmlReader.Create(new System.IO.StringReader(xsdText))
        schemaSet.Add(null, reader) |> ignore
        schemaSet.Compile()
        schemaSet


    [<Literal>]
    let AttributeWithDefault = """
    <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
        <xs:element name="test">
            <xs:complexType>
                <xs:attribute name="myattr" use="optional" type="xs:string" default="a default value"/>
            </xs:complexType>
        </xs:element>
    </xs:schema>"""

    let doc = XDocument.Parse """<test />"""
    let xsd = parseSchema AttributeWithDefault

    doc.Root.Attribute(XName.Get "myattr") |> printfn "%A" // <null>
[...]

PR #1209, opened 2018-09-14
Add GetSchema method in XML type provider

giacomociti:
This is related to #1189 .

A method `GetSchema` is added (much like `GetSample`, and in fact the implementation is a hack to make schema parsing fit in code intended to load samples).

Once the schema is available (as a `System.Xml.Schema.XmlSchemaSet` object) it can be used to validate input like this:

    type T = XmlProvider<Schema="....">
    let s = T.GetSchema()
    let d = T.Parse("...")
    d.XElement.Document.Validate(s, null)

The `Validate` method has overloads to customise the behaviour (like applying default values as requested in #1189 ).

I'm not sure this is worth merging because, as already stated, the implementation is a hack.


ovatsus:
Can you update the documentation explaining how to use GetSchema, and when to use it?

ovatsus:
Thanks

Issue #1221, opened 2018-10-23
Numeric inference in 3.0.0 appears broken


panesofglass:
In FSharp.Data 3.0.0-beta4, inference for a JSON with a numeric array field worked properly. When attempting to upgrade to 3.0.0, however, the inference breaks, and the inferred type is `int []`. I'm currently working on a repro test case. This is preventing our ability to update to the 3.0.0 release version.

panesofglass:
I suspect this commit may be the culprit, but I'm still looking: https://github.com/fsharp/FSharp.Data/commit/8d05e4bf32fe750cb0f1bad977cbd5b97b39ab17

atlemann:
What does your numeric array field look like? After bumping to v3.0, `0.1e1` is inferred as int. To fix it I had to change to `0.1e0` or `0.11e1`. 

For arrays, make sure all the entries are in exponential format. Maybe that helps.

panesofglass:
I'm figuring out ways to work around this, but I think the general inference should be that, if there's an `e` or `E` in the string representation, it should be a `float`.

baronfel:
I think we could solve this with some additional information attached to the `JsonValue.Float` case that let us know if the float value was exponential or not.  Then we could use that flag to disallow inferring the sample value as an integer at all.

dsyme:
Actually looking at this I think we don't change this.  All the solutions are too invasive

PR #1222, opened 2018-10-23
Fix incorrect JsonProvider numeric inference

panesofglass:
See #1221 

Hypothesis is that commit 8d05e4bf32fe750cb0f1bad977cbd5b97b39ab17 is not quite correct.


tests/FSharp.Data.Tests/JsonProvider.fs
```diff
@@ -734,3 +734,18 @@ let ``Getting a large float at runtime when an integer was inferred should throw
     let f = 1234567890123456789012345678901234567890.
     let json = JsonProvider<"""{ "x" : 0.500, "y" : 0.000 }""">.Parse("""{ "x" : -0.250, "y" : 1234567890123456789012345678901234567890 }""")
     (fun () -> json.Y) |> shouldThrow (sprintf "Expecting an Int32 at '/y', got %s" (f.ToString("E14").Replace("+0", "+").Replace(",", ".")))
+
+[<Test>]
+let ``Getting field with numeric values should infer decimal``() =
+    let json = JsonProvider<""" { "a":[1.0, 1.0] } """>.GetSample()
```
panesofglass: I _think_ this is supposed to infer `decimal`, but I could be wrong.

panesofglass:
Appears these lines are the issue: https://github.com/fsharp/FSharp.Data/blob/8d05e4bf32fe750cb0f1bad977cbd5b97b39ab17/src/Json/JsonInference.fs#L34-L35

What's wanted here? I would expect the format of the number to override the use of `Math.Round(v) = v` as a means of inference. If not, what is required to make this work correctly?

panesofglass:
@ovatsus @baronfel I don't know the right way to fix this. Help is appreciated.

baronfel:
Personally I think the single flag for `inferTypesFromValues` is a bit wide for coercing both strings _and_ numeric types.  I'd be in favor of expanding the number of static/configuration parameters for this (maybe just a split for string-inference and number-inference) as well as expanding the number of JsonValue cases to include `Integer` and `LongInteger` or something semantically similar, since the only time where we have enough information is at JsonValue.Parse time. Armed with JsonValue.Integer and JsonValue.LongInteger we could easily handle type mapping to int/int64 for json provider even without `inferTypesFromValues` set.

dsyme:
Close/open to re-run CI

dsyme:
Closing this out as the repo has diverged a long way since the proposal was made

panesofglass:
We ultimately never upgraded and are stuck on FSharp.Data 3.3.3.

PR #1224, opened 2018-10-26
Failing test case for #1223 - Content from tags inside pre is missing whitespace characters

### LINKED ISSUE #1223: Content from tags inside pre is missing whitespace characters
When emitting a parsed document whitespace characters are usually preserved for `pre` tags. The following test checks this.

https://github.com/fsharp/FSharp.Data/blob/33e6e825bc2978eb9ce6dd880f31d9e60d452699/tests/FSharp.Data.Tests/HtmlParser.fs#L763-L772

However that's not the case when whitespace characters are within a tag that's within the `pre` tag, so e.g. when emitting the parsed document from the following snippet, whitespace characters are missing.

```
<pre><span>\r\n        This code should be indented and\r\n        have line feeds in it</span></pre>
```

The code that's responsible for "normalizing" whitespace characters is the following:

https://github.com/fsharp/FSharp.Data/blob/33e6e825bc2978eb9ce6dd880f31d9e60d452699/src/Html/HtmlParser.fs#L373-L382

And `x.InsertionMode` is calculated as follows:

https://github.com/fsharp/FSharp.Data/blob/49a3bfb22a8955463d7536af1d2df86449e335c6/src/Html/HtmlParser.fs#L353-L356

`x.IsFormattedTag` is only `true` if the current node is `pre` or `code`, but I think it should also check its ancestors if it's somewhere _within_ a `pre` tag, shouldn't it?

johannesegger:
see #1223. This PR currently only adds the failing test, because I have no clue how to fix it.


johannesegger:
I have the same problem when building locally as AppVeyor has, i.e. FSharp.Formatting can't be downloaded, any ideas?

dsyme:
Close/open to re-run CI

johannesegger:
Thanks. Now it's showing the failing test. Do you agree this is a bug?

dsyme:
Yes, agreed this is a bug

cartermp:
Closing old pr

PR #1226, opened 2018-11-05
NameUtils/Pluralizer: Fix unusual singularizations for some common words (choices; releases)

rmuch:
Currently, "releases" takes the "sis" <-> "ses" suffix rule, and therefore becomes "releasis" when converted from plural to singular.

This patch fixes the problem by adding "release" as a special word, so that `toSingular` correctly resolves the singular form.

### DIFF
--- a/src/CommonRuntime/Pluralizer.fs
+++ b/src/CommonRuntime/Pluralizer.fs
@@ -98,6 +98,7 @@ let private tables = lazy(
          "child",            "children",         ""
          "chassis",          "",                 ""
          "chinese",          "",                 ""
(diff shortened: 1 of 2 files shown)


rmuch:
Noticed another weird one today with "Choices" being singularized to "Choicis". Will update the pull request with another special word. I wonder how many more common words that have this problem?

dsyme:
Close/open to re-run CI

dsyme:
@ovatsus This seems legit, merging to be in 3.0.3

cartermp (approved):
Thank you for this.

Issue #1215, opened 2018-10-05
Support JSON comments


mjmckp:
JSON.Net supports both `//` and `/* .... */` style comments (https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Linq_CommentHandling.htm), can FSharp.Data be extended to do the same?

PR #1228, opened 2018-11-07
Addresses: Support JSON comments #1215

MaheshPunna:
Previous behavior:
The Json type provider fails with invalid character if the json file contains comments for the json fields
Current behavior:
this commit recognizes comments for the json fields and ignores them

### DIFF
--- a/src/Json/JsonValue.fs
+++ b/src/Json/JsonValue.fs
@@ -155,6 +155,23 @@ type private JsonParser(jsonText:string) =
     let skipWhitespace() =
       while i < s.Length && Char.IsWhiteSpace s.[i] do
         i <- i + 1
+    let skipLine() =
+        while i < s.Length && not (s.[i] = '\n') do
+            i <- i+1
+        i <- i+1
+        skipWhitespace()
+
+    let skipMultiLineComment()=
+        while i < s.Length && not (String.Equals(s.[i..i+1],"*/")) do
+            i <- i+1
+        i <- i+2
+        skipWhitespace()
+
+    let skipComments()=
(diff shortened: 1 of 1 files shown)


eugbaranov:
I think it begs for extra unit tests for cover for possible occurrence of comments in various places e.g. before/after/inside name and values.
Also although admittedly I only glanced over commit but how will it handle if there is a whitespace after _comment_? What if there multiple interleaving spaces and comments?

src/Json/JsonValue.fs
```diff
@@ -155,6 +155,23 @@ type private JsonParser(jsonText:string) =
...
       while i < s.Length && Char.IsWhiteSpace s.[i] do
         i <- i + 1
+    let skipLine() =
+        while i < s.Length && not (s.[i] = '\n') do
+            i <- i+1
+        i <- i+1
+        skipWhitespace()
+
+    let skipMultiLineComment()=
+        while i < s.Length && not (String.Equals(s.[i..i+1],"*/")) do
```
eugbaranov: Will this not go boom 💥 if `i` gets to `s.Length-1`?

MaheshPunna: Hi,
Case-1: white spaces after comments are handled by skipwhitespaces(), that is called after a new line character is found
Case-2: for a single line comment, the terminating condition is new line character, so it skips until it finds that. 
If i=s.length-1 and is still looking for coment terminating chars, then it's anyway wrong JSON format,

Thanks for the comments, I will add test cases for all these suggestions

dsyme:
Close/open to re-run CI

dsyme:
Closing old pull request, please resurrect if possible

Issue #1238, opened 2019-01-02
CSV changes to enable Markdown Table Parsing


xdaDaveShaw:
I had a thought the other day (OK a while ago), that it would be nice to work with a Markdown Table as if it was CSV Data.

```
| Tables        | Are           | Cool  |
| ------------- |:-------------:| -----:|
| col 3 is      | right-aligned | $1600 |
| col 2 is      | centered      |   $12 |
| zebra stripes | are neat      |    $1 |
```

The Pipe `|` separator doesn't really work, there's a couple of things that get in the way.

Instead of adding just a "Markdown Mode" here are 2 suggestions that enable you to parse Markdown Tables with the CSV Provider.

- Ignore Line Pattern- a Regex pattern that if met means a line of the CSV file is ignored.
This could be used to stop it looking at the line between the headers and the content.

- Trim Column Value (yes/no)
This would call `String.Trim` on the call value before returning it. It also helps in the type inference get the correct types.

I've started work on this, and am just merging with the latest and polishing it off before I think about submitting a PR, does anyone see any value in this?

PR #1239, opened 2019-01-02
New functions that would allow Markdown Tables

xdaDaveShaw:
Implementation of #1238 

Two new options on the `CsvProvider` type:

- `IgnoreLinePattern`: A regular expression for lines to ignore.
- `TrimColumnValue`: Whether the values in a column should be trimmed.


xdaDaveShaw:
This is my first PR to this repo and is also my first work on Type Providers - and it is only an idea - so feel free to reject it, or ask for improvements.

dsyme:
Close/open to re-run CI

xdaDaveShaw:
Looks like I need to fix some things - will take a look tonight if I have time

dsyme:
@xdaDaveShaw Apologies for the late reply.

This is interesting - though should there be a separate "MarkdownProvider"?   It seems odd to include this as part of CSV.

xdaDaveShaw:
@dsyme - I was only referring to Markdown Tables - not all of Markdown - that might be more of a challenge.

The have an "almost" CSV like format, if only the header/rows separator could be ignored, and values trimmed to fix the type inference. 

e.g. 

```
|Col1|Col2|
|----|----|      <- Skipping this...
|1   |str |
  ^^^            <- trimming this
```

I don't mind this been rejected if there are no other uses for the feature and Markdown would be better served elsewhere.

dsyme:
Closing this out due to repo divergence.  Please resubmit if you'd like to

Issue #1242, opened 2019-01-28
CsvFile.Save doesn't delete existing file with the same name before writing?


nhirschey:
I'm not sure if this is by design, but it appears that if I call CsvFile.Save(file) where file is an existing file, the type provider will not delete the existing file before writing. Instead, it overwrites it.  This is a serious problem if the new CsvFile is smaller than the prior one, resulting in corrupted CSVs.

I think it makes much more sense to first delete an existing file with the same name, then write to that file name. Is there any reason to not first delete a pre-existing file with the same name?

https://github.com/fsharp/FSharp.Data/blob/de2975c85299b805d5a72ea3eac1519f464c802a/src/Csv/CsvRuntime.fs#L356

nhirschey:
Providing an example of the problem

```fsharp
type MyCsvType =
  CsvProvider<Schema = "A (int), B (string), C (date option)",
    HasHeaders=false>

let myRows =
  [ MyCsvType.Row(1, "a", None)
    MyCsvType.Row(2, "B", Some DateTime.Now) ]

let file = """C:\test.csv"""
let myCsv = new MyCsvType(myRows)
myCsv.Save(file)

MyCsvType.Load(file).Rows
|> Seq.length 
// val it : int = 2

// Close fsi so that type provider releases file

type MyCsvType =
  CsvProvider<Schema = "A (int), B (string), C (date option)",
    HasHeaders=false>

let myRows =
  [ MyCsvType.Row(1, "a", None)
    MyCsvType.Row(2, "B", Some DateTime.Now) ]

let file = """C:\test.csv"""

let myCsv2 = new MyCsvType(Seq.take 1 myRows)
myCsv2.Save(file)

MyCsvType.Load(file).Rows
|> Seq.length
// val it : int = 2, but should be 1
```

nhirschey:
Fixed by https://github.com/fsharp/FSharp.Data/pull/1243

PR #1243, opened 2019-01-29
Fix Csv.Save() Data Corruption trap: If x.Save(file) is called and file already exists, delete it and then…

nhirschey:
This is to fix the unexpected behavior in this issue: https://github.com/fsharp/FSharp.Data/issues/1242

I can't think of a scenario where you would want to (partially) write over a pre-existing file that you are calling x.Save() on. I found this investigating why csv files I was creating were corrupted with bad data after I changed data and tried resaving the resulting csv. 

I also think this behavior is more consistent with csv saving features of other languages people using this library might be coming from (SAS, R, Stata, etc.).


dsyme:
Close/open to re-run CI

nightroman:
Would not it be better to open the file with the "override" option instead of deleting? Presumably, this way is more effective preserves file permissions, too.

nhirschey:
I do not know what the "override if exists" code is. I just know that without this fix, file.Save("...") corrupts data if the new csv file has fewer rows than the original that it is replacing.

If there's a more common code conventionfor replacing a file if it exists, I'm fine with that.

nightroman:
I think it's [FileStream(String, FileMode, FileAccess, FileShare)](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.-ctor?view=netframework-4.7.2#System_IO_FileStream__ctor_System_String_System_IO_FileMode_System_IO_FileAccess_System_IO_FileShare_) with `FileMode` set to `Create`.

nightroman:
or this `File.Open(<path>, FileMode.Create, FileAccess.Write, FileShare.None)) ...`

nightroman:
I think the simplest way is to replace the original line

    let writer = new StreamWriter(File.OpenWrite(path))

with

    let writer = new StreamWriter(path)

This constructor uses `append = false`, i.e. the existing file is overridden.
That is how Deedle writes CSV, see
https://github.com/fslaborg/Deedle/blob/f0d3a2e61268e9d666b1f95cd371e135aa74c071/src/Deedle/FrameExtensions.fs#L641

nhirschey:
Thanks for the link to Deedle’s version. I’ll update to that, and the Deedle reference will help justify the pull request.

nhirschey:
The new version follows the example of Deedle to ensure data corruption does not occur when a user calls CsvFile.Save(path) where path is an existing file on disk.

As an example of the new functionality is below. See the referenced issue for the behavior in the current released version. 

```fsharp
#r """packages\FSharp.Data\lib\net45\FSharp.Data.dll"""
open FSharp.Data
open System

type MyCsvType =
  CsvProvider<Schema = "A (int), B (string), C (date option)",
    HasHeaders=false>

let myRows =
  [ MyCsvType.Row(1, "a", None)
    MyCsvType.Row(2, "B", Some DateTime.Now) ]

let file = __SOURCE_DIRECTORY__ + "/test.csv"
let myCsv = new MyCsvType(myRows)
myCsv.Save(file)

IO.File.ReadAllLines(file)
// Here is the original file
//val it : string [] = [|"1,a,"; "2,B,2019-04-16T11:34:37.5933828+01:00"|]

type MyCsvType2 =
  CsvProvider<Schema = "A (string), B (string), C (date option)",
    HasHeaders=false>

let myRows =
  [ MyCsvType2.Row("blah", "a", None) ]

let myCsv2 = new MyCsvType2(Seq.take 1 myRows)
myCsv2.Save(file)

IO.File.ReadAllLines(file)

// This version correctly only has one line in the resulting file.
// val it : string [] = [|"blah,a,"|]
```

PR #1248, opened 2019-03-19
Fixed undisposed writers in CSV saving

daniel-chambers:
This fixes a small bug where the CSV Save functions don't dispose the writers they use, and therefore leave stream flushing and potentially file handles open until garbage collection cleans up the writers. I've simply changed the `let writer`s to `use writers` to ensure they are disposed correctly.

One point that's worth some discussion: for the `Save(stream : Stream ...` overload, I've made it so that the save does not close the underlying stream. Instead, the writer flushes into it, but leaves it open. I chose to do this because I'm of the opinion that if I give a stream to something, I (the caller) own it and I will close and dispose it when I'm done with it. I don't expect `Save` to close and dispose it for me.
However this is not the default behaviour when you pass a Stream to a Writer, hence why I'm using a more complicated overload.

### DIFF
--- a/src/Csv/CsvRuntime.fs
+++ b/src/Csv/CsvRuntime.fs
@@ -347,18 +347,18 @@ type CsvFile<'RowType> private (rowToStringArray:Func<'RowType,string[]>, dispos
           writer.Write item)
 
   /// Saves CSV to the specified stream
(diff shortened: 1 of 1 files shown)


dsyme:
Close/open to re-run CI

Issue #1253, opened 2019-03-28
Possible problem with providedTypesCache


dsyme:
I'm skeptical about the validity of the `providedTypesCache` approach in FSharp.Data when type provider instances get disposed and re-created.

This will be returning ProvidedTypeDefinition from previous TP instances which may in turn hold handles to those TP instances, artificially extending the lifetime of those TP instances and the resources they use, at the moment for up to 5 minutes - over this time a lot of data can be leaked if TP instances are not released correctly 

I think I would prefer that ProvidedTypeDefinition objects were not cached in global tables.

@ovatsus Do you remember the reasoning here? Is this solving a dramatically important performance issue?  

thanks
don

PR #1254, opened 2019-03-29
decrease cache time

dsyme:
I'm convinced 5 minutes is too long to be caching these instances under many user edits which might recreate many type provider instances

I'm also sceptical about the cache in general, see #1253 - I'd much prefer this is scoped to a particular TP instance - but for now decreasing the cache time seems sensible.

This is all part of a range of work to try to reduce the impact of some memory problems we're seeing in when using F# tooling with FSharp.Data type providers


dsyme:
@baronfel @ovatsus Any chance you could push FSHarp.Data 3.0.1 after this gets merged? Thanks

Issue #1261, opened 2019-05-01
json root values


graydon:
Hi,

It looks like in https://github.com/fsharp/FSharp.Data/issues/563 and its patch https://github.com/fsharp/FSharp.Data/pull/594 the root value of a Json text was restricted to just objects and arrays, based on [this comment](https://github.com/fsharp/FSharp.Data/issues/563#issue-31863788) suggesting that both jsonlint and the spec do not support general values at the root.

As far as I can tell this was based on a state of affairs that has subsequently changed, specifically [RFC 4627 from 2006]( https://tools.ietf.org/html/rfc4627) in which such a restriction held.

Subsequently this restriction was removed, and non-object-and-array values (eg. null, true, false) are all allowed at root. [RFC7195 in 2014](https://tools.ietf.org/html/rfc7159#section-2) and [RFC8259 in 2017](https://tools.ietf.org/html/rfc8259#page-5) and [ECMA-404 from 2017](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) and the revised grammar on [json.org](https://json.org/) and [jsonlint](https://jsonlint.com/) all agree on this point presently.

(I'm not just being a pedant here -- I have an application that produces such values and this parser breaks on it)

graydon:
Opened PR #1262 above.

PR #1262, opened 2019-05-01
Remove obsolete special-case root value parsing.

graydon:
Quick stab at fixing #1261. Seems to work now with the second fix applied.


graydon:
Updated with fix for incorrect assert in parseLiteral.

dsyme:
Great work!

Issue #1255, opened 2019-04-01
Build with embedded resource broke after upgrade from 3.0.0 to 3.0.1 


Karamell:
I get an error message if I upgrade JsonProvider from 3.0.0 to 3.0.1:

Error: 
```
error FS3033: The type provider 'ProviderImplementation.JsonProvider' reported an error: The operation 'reading manifest resource %s' on item 'lib.file1.json' should not be called on provided type, member or parameter of type 'ProviderImplementation.ProvidedTypes.TargetAssembly'. Stack trace:
   at System.Environment.get_StackTrace()
   at ProviderImplementation.ProvidedTypes.Utils.notRequired[a,b](a this, String opname, String item) in C:\Git\FSharp.Data\paket-files\fsprojects\FSharp.TypeProviders.SDK\src\ProvidedTypes.fs:line 169
   at ProviderImplementation.ProviderHelpers.readResource(DisposableTypeProviderForNamespaces tp, String resourceName) in C:\Git\FSharp.Data\src\CommonProviderImplementation\Helpers.fs:line 177
   at ProviderImplementation.ProviderHelpers.parseTextAtDesignTime@198.Invoke(IDisposable _arg1) in C:\Git\FSharp.Data\src\CommonProviderImplementation\Helpers.fs:line 210
   at Microsoft.FSharp.Core.Operators.Using[T,TResult](T resource, FSharpFunc`2 action)
   at ProviderImplementation.ProviderHelpers.generateType@368.Invoke(Unit unitVar0) in C:\Git\FSharp.Data\src\CommonProviderImplementation\Helpers.fs:line 377
   at ProviderImplementation.ProviderHelpers.getOrCreateProvidedType@293-3.Invoke(IDisposable _arg1) in C:\Git\FSharp.Data\src\CommonProviderImplementation\Helpers.fs:line 341
   at Microsoft.FSharp.Core.Operators.Using[T,TResult](T resource, FSharpFunc`2 action)
   at <StartupCode$FSharp-Data-DesignTime>.$ProvidedTypes.clo@9061-164.Invoke(FSharpFunc`2 f, String s, Object[] p) in C:\Git\FSharp.Data\paket-files\fsprojects\FSharp.TypeProviders.SDK\src\ProvidedTypes.fs:line 9063
[...]

PR #1273, opened 2019-07-04
Fix issue #1255 : EmbeddedResource not working

fpellet:
I take old version of FSharp.TypeProviders.SDK

Fix issue #1255


dsyme:
Hi @fpellet, We need to get to the bottom of this issue rather than reverting to the old TPSDK.  

Do you happen to know the root change to the TPSDK that caused problem?

fpellet:
Hi @dsyme 
I think it's the next version that I locked. After, I did not find the exact problem source. This isn't the easiest function to understand :)

dsyme:
So perhaps this change: https://github.com/fsprojects/FSharp.TypeProviders.SDK/commit/dff4ef5d9e0f6130586326069bef25efc64c427f#diff-7c540ae6a67f14c48d0403e7b021fc86L7954

fpellet:
The last version fix this bug  
I rebase and rollback commit 965c2d1  
Now, this PR add only a test to detect regression

dsyme:
Thank you!!!

Issue #1259, opened 2019-04-17
FSharp.Data emits invalid JSON for NaN, Infinity and -Infinity


lkmfwe:
FSharp.Data serializes NaN as just `NaN`, Infinity as just `Infinity` and -Infinity as `-Infinity`.
For example, Firefox console encodes these values like this:
```
>> JSON.stringify([NaN, Infinity, -Infinity])
<< "[null,null,null]"
```
If I understand it correctly, there is no JSON representation for these values, so emitted JSON is not valid.
Repro script:
```fsharp
open FSharp.Data

let serializeDeserializeFloat floatValue = 
    let jsonValue = JsonValue.Array ([| JsonValue.Float floatValue |])
    
    let json = jsonValue.ToString(JsonSaveOptions.DisableFormatting)
    printfn "Serialized: %A" json
    
    let jsonValue = JsonValue.Parse json
    printfn "Deserialized: %A" jsonValue
    
serializeDeserializeFloat 42.0
// OK
serializeDeserializeFloat nan    
// Throws exception
// System.Exception: Invalid JSON starting at character 1, snippet = 
// ----
// [NaN]
// -----
// json = 
// ------
// [NaN]
// -------
serializeDeserializeFloat infinity
//System.Exception: Invalid JSON starting at character 1, snippet = 
//----
//[Infinity]
//-----
//json = 
//------
//[Infinity]
//-------
serializeDeserializeFloat -infinity
//System.Exception: Invalid JSON starting at character 2, snippet = 
//----
//[-Infinity]
//-----
//json = 
//------
//[-Infinity]
//-------
```

lkmfwe:
For example Newtonsoft.Json serializes these values as strings. NaN is `"NaN"`, Infinity is `"Infinity"` and -Infinity is `"-Infinity"`, so it is able to deserialize it back to float and JSON is valid.
```fsharp
open Newtonsoft.Json

let serializeDeserializeFloatNewtonsoft floatValue =
    let json = JsonConvert.SerializeObject([| floatValue |])
    printfn "Serialized %A" json
    let deserialized = JsonConvert.DeserializeObject<float[]>(json)
    printfn "Deserialized %A" deserialized
    
serializeDeserializeFloatNewtonsoft 20.0
serializeDeserializeFloatNewtonsoft nan    
serializeDeserializeFloatNewtonsoft infinity
serializeDeserializeFloatNewtonsoft -infinity
```
Output:
```
Serialized "[20.0]"
[...]

PR #1275, opened 2019-07-24
Handle serialization of odd float values

mjgpy3:
Unfortunately `System.Double.NaN`, `System.Double.PositiveInfinity` and `System.Double.NegativeInfinity` cause invalid JSON to be generated. These seem like special cases that shouldn't happen often but they are not disallowed by the type system.

I was using FsCheck to generate `JsonValue` instances and I noticed an array containing `NaN` which causes JSON parsers to error.

This change attempts to mimic the decisions already made in JavaScript, e.g.
``` javascript
$ node
> JSON.stringify(NaN)
'null'
> JSON.stringify(Infinity)
'null'
> JSON.stringify(-Infinity)
'null'
```

### DIFF
--- a/src/Json/JsonValue.fs
+++ b/src/Json/JsonValue.fs
@@ -63,10 +63,10 @@ type JsonValue =
       else "\":"
 
     let rec serialize indentation = function
-      | Null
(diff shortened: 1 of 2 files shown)


mjgpy3:
I just looked through the issues and it appears that this resolves #1259

PR #1278, opened 2019-07-31
Refactor parse' to use tail recursion

markfields:
There's one non-tail-recursive call, which is the one when stepping into a tag
to parse its children.  This change adds a manual callstack to the function
onto which we push the variables in the closure before continuing to the
child tokens.  Then in the places where we were returning, call a helper
to pop the stack and include the logic that originally followed the recursive call.

This should represent no change in behavior, only a resiliency against
StackOverflowException.


markfields:
I forked to the wrong repo, which I since deleted.  I'll resubmit after forking properly.

Issue #1277, opened 2019-07-30
StackOverflowException thrown from HtmlDocument.Load


markfields:
We're seeing StackOverflowException repeatedly in our production service with our HTML parsing code implicated.  I managed to find a repro that hit StackOverflowException and debugged it to find that the line that throws in that case is a call to `FSharp.Data.HtmlDocument.Load`.

Taking a quick look at the code here, I found a non-tail-recursive call [here ](https://github.com/fsharp/FSharp.Data/blob/49a3bfb22a8955463d7536af1d2df86449e335c6/src/Html/HtmlParser.fs#L810) in `parse'` under `HtmlDocument.Load`.  Next step will be to download the source code and load symbols for `FSharp.Data` and see what the callstack is.

The test case that triggered this in our production-like test environment is a `p` tag containing 4090 nested `span` tags:
[4090depth.txt](https://github.com/fsharp/FSharp.Data/files/3448523/4090depth.txt)

This doesn't repro in our dev environment, so there may be some runtime configuration in prod that adjusts the max stack size?  At any rate, using tail recursion everywhere will preclude this kind of bug altogether.

markfields:
I've made (but not yet tested beyond building) a minimally invasive fix, you can see it in #1278 but I need to resubmit the PR from a different fork, so I closed that one.

cartermp:
I will close this since https://github.com/fsprojects/FSharp.Data/pull/1279/ has been merged. Thanks for the contribution!

PR #1279, opened 2019-08-01
Refactor parse' to use tail recursion

markfields:
This should bring no change in behavior, only a resiliency against `StackOverflowException` (see #1277).

There's one non-tail-recursive call in `parse'`, which is the one when stepping into a tag
to parse its children. This change adds a manual callstack to the function
onto which we push the variables in the closure before continuing to the
child tokens. Then in the places where we were returning, call a helper
to pop the stack and include the logic that originally followed the recursive call.

All other recursive calls are already tail recursive, and represent a go-to-next-sibling step rather than a go-to-first-child step in the resulting HTML.  So they remain unchanged.


.gitignore
```diff
@@ -23,4 +23,5 @@ paket-files
 TestResults
 *.log
 .vscode
-.idea
\ No newline at end of file
+.idea
+.ionide/
```
markfields: I can revert this change if requested; it was necessary in my dev environment where I use the ionide plugin for VS Code.

markfields:
@ovatsus care to take a look at this change?

I haven't been able to sort out the Travis failure, and I'm also unable to build/run the tests locally, so I wouldn't merge until I get that figured out.  But looking for feedback on the concept/implementation.  Thanks.

dsyme:
Reopen for travis

markfields:
Wow forgot about this. Not sure what to make of all the timeout errors in the AppVeyor test run. @dsyme do you suppose they're related to this change or not? Skimming the callstacks I don't see anything that obviously would be doing HTML Parsing. Obviously if I wrote an infinite loop that would cause timeouts though... :)

cartermp:
@markfields sorry for no response here - we've reworked several aspects of this repository and I'm scanning through contributions that can be taken in now. This one looks good, subject to tests passing. No more AppVeyor :)

markfields:
Thanks @cartermp!

PR #1281, opened 2019-08-15
Fix StackOverflow exception caused by many ampersand and semicolins i…

berickson1:
…n attribute values

The original attributeValueCharRef function took a continutation parameter that was always a reference back to the caller. Splitting this into two functions (one where each case is handled) removes the non-tail recursive call.

This can occur if an html attribute has a bunch of encoded html inside of it


berickson1:
Going to complete this PR in VSTS https://o365exchange.visualstudio.com/O365%20Core/_git/Griffin/pullrequest/455969 and sync back to github instead of the reverse

dsyme:
@berickson1 Could you add a test case for this?

berickson1:
To do this I'm going to have to enable tail calls in FSharp.Data debug builds - since tail recursion is what prevents this from crashing. I'll update the PR to include this and let you decide if it's desirable or not

dsyme:
Thanks for adding the test case!!

PR #1284, opened 2019-09-04
Create a new ToString function

precisemath:
Create a new ToString function with an additional parameter to change the culture Info.


dsyme:
Could you add a test case for this?

precisemath:
@dsyme when creating the test cases I realized that we dont really need this function as the culture change doesnt really affect int, float and string. It mainly affects datetime. So closing this Pull Request.

PR #1286, opened 2019-09-10
Update to latest TPSDK

dsyme:
Fixes a problem with using FSHarp.Data on .NET Core F# Interactive


dsyme:
I have ignored a failing XML test that is something to do with either network connectivity or a Mono update

PR #1297, opened 2020-01-02
Fix cookies to be added in cookieContainer when it is passed as argument

devcrafting:
I intended to do so as explained in comment of https://github.com/fsharp/FSharp.Data/commit/5fdc7fdbbe4257847250ad5083142478c00de961, but the boolean was reversed.

It worked most of the time because dotnet framework take care of adding cookies based on Set-Cookie header most of the time. But it fails if the cookie Path does not match the ResponseUri. It was a requirement of https://tools.ietf.org/html/rfc2965 et https://tools.ietf.org/html/rfc2109, but the last version (still in "proposed standard") https://tools.ietf.org/html/rfc6265 does enforce Domain match only, no more path. I scrapped a website sending a Set-Cookie where Path was not matching ResponseUri, and it works smoothly in all browsers, so I think we can rely on last RFC version here.

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1536,8 +1536,8 @@ type Http private() =
         // set cookies
         let addCookiesFromHeadersToCookieContainer, cookieContainer =
             match cookieContainer with
-            | Some x -> false, x
-            | None -> true, CookieContainer()
+            | Some x -> true, x
+            | None -> false, CookieContainer()
 
         match cookies with
         | None -> ()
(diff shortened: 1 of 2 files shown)


devcrafting:
Seems AppVeyor is failing because of "System.Exception : Cannot read sample XSD from 'http://europa.eu/rapid/conf/RSS20.xsd': The remote server returned an error: (404) Not Found." not related to the modification, any ideas ?

dsyme:
> Seems AppVeyor is failing because of "System.Exception : Cannot read sample XSD from 'http://europa.eu/rapid/conf/RSS20.xsd': The remote server returned an error: (404) Not Found." not related to the modification, any ideas ?

I guess disable the test, either in this PR or as a separate PR - thank you!

cartermp:
closing old pr

PR #1304, opened 2020-02-04
Add DateTimeOffset for Csv Schema Inference

nhirschey:
The goal is to allow specifying 
```fsharp
type MyCsvType =
  CsvProvider<Schema = "A (int), B  (datetimeoffset option)",
    HasHeaders=false>
```
The changes are in lines 22, 32, and 41. The rest of the differences are whitespace edits to keep formatting consistent with the prior style of lining everything up vertically.


NickDarvey:
What's required to get this merged?

nhirschey:
The maintainers probably want to chime in on the Date schema API. The default was changed date for the provider was changed to DateTimeOffset in the 3.X move, but none of the schema code was updated to correspond. 

An alternative to my changes here would be to change the old code so that 'date' implies DateTimeOffset rather than DateTime. I think having the `date` schema imply DateTimeOffset might make more sense given the current DateTimeOffset default, but that causes backwards compatibility issues and was forgotten in the 2.x (DateTime default) -> 3.X  (DateTimeOffset) move.

dsyme:
> What's required to get this merged?


The change looks fine, can you add test cases for this?

nhirschey:
Will do.

________________________________

nhirschey:
I think this is now ready.

Additions:
- Add DateTimeOffset to structural types, which was required to get inference working correctly.
- I added a test to make sure that the schema inference was working correctly, for example when parsing `name (datetimeoffset option)` and `name (datetimeoffset?)`. 
- I updated the schema inferencing documentation for CsvProvider to account for the changes.

This set of commits depends on another pull request (https://github.com/fsharp/FSharp.Data/pull/1320), which has changes necessary to get the build working.

PR #1316, opened 2020-06-02
Modified escaping of query string parameters

### LINKED ISSUE #1263: Inadequate escaping of query parameters
Another request-to-revert: I spent the day tracking down an inadequate level of escaping in query parameters (in my case, `+` characters in query parameters that pass through `EscapeUriString` unescaped, then get un-URL-escaped on the receiving side as spaces) due to https://github.com/fsharp/FSharp.Data/issues/1137. Since `Http.Request` _does_ pass its input through `EscapeUriString`, I can't exactly work around it by escaping myself -- the `%2F` I would substitute itself gets escaped to `%252F`. So I'm actually having to avoid using the whole `Http.Request` subsystem because of this.

`EscapeDataString` may make ugly URLs, but they're at least correctly escaped. Could `Http.Request` be changed back to use it?

paulgartner1:
fixes #1263 

Escape query strings using `Uri.EscapeDataString` instead of `Uri.EscapeUriString`, this should correctly escape all URI reserve characters so they can be used as literals in query strings.

- Updated failing test
- Added specific unit test

### DIFF
--- a/tests/FSharp.Data.DesignTime.Tests/SignatureTestCases.config
+++ b/tests/FSharp.Data.DesignTime.Tests/SignatureTestCases.config
@@ -33,7 +33,6 @@ Xml,TypeInference.xml,false,false,,false,
 Xml,SampleAzureServiceManagement.xml,false,false,,true,
 Xml,TimeSpans.xml,false,false,,true,
 Xml,,false,false,,false,po.xsd
-Xml,,false,false,,false,http://europa.eu/rapid/conf/RSS20.xsd
 Xml,,false,false,,false,homonim.xsd
 Xml,,false,false,,false,IncludeFromWeb.xsd
 Json,WorldBank.json,false,WorldBank,,true
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1505,11 +1505,6 @@ type Http private() =
     static member internal EncodeFormData (query:string) =
         (WebUtility.UrlEncode query).Replace("+","%20")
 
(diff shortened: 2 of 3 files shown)


dsyme:
Thanks!  This looks like exactly the fix we need.

@cartermp If you get a moment could you review?

ovatsus:
EscapeDataString has problems with large payloads, that's why we changed it, see #1112

ovatsus:
Ah, nevermind, that was for the body, which you are not changing

tests/FSharp.Data.DesignTime.Tests/SignatureTestCases.config
```diff
@@ -33,7 +33,6 @@ Xml,TypeInference.xml,false,false,,false,
 Xml,SampleAzureServiceManagement.xml,false,false,,true,
 Xml,TimeSpans.xml,false,false,,true,
 Xml,,false,false,,false,po.xsd
-Xml,,false,false,,false,http://europa.eu/rapid/conf/RSS20.xsd
```
ovatsus, line 36: Was this removed by accident?

paulgartner1: I removed it as `RSS20.xsd` does not exist at url anymore and was causing unit tests to fail.

I could not find an updated url.
Appologies I should have updated the PR comment.

dsyme: That URL is now retuning a failure, I think it can be removed

Issue #1191, opened 2018-08-20
Error opening a library that uses JsonProvider with an embedded resource


ScottShingler:
I've encountered some strange behaviour when trying to open a library that uses a JsonProvider with an embedded resource. 

Given then following module defined in a separate library:

```fsharp
module MyLib

open FSharp.Data

type MyJsonConfig = JsonProvider<"sample.json", EmbeddedResource="MyLib, sample.json">

type MyFoo = { Foo : string }

let loadMyConfig (path : string) =
    let config = MyJsonConfig.Load(path)
    { Foo = config.Foo }
```

And the following program that references and consumes `MyLib`:

```fsharp
open System
open MyLib

[<EntryPoint>]
let main argv =
    let config = loadMyConfig("config.json")
    printfn "Loaded config: %A" config
    printfn ""
    printfn "Press any key to exit"
    Console.ReadKey() |> ignore
    0
```

I get the following compilation error:

> Error FS3033 The type provider 'ProviderImplementation.JsonProvider' reported an error: An index satisfying the predicate was not found in the collection.

For this simple example, there is a workaround to avoid this error: Instead of opening `MyLib`, use the fully-qualified name when referencing `loadMyConfig` (i.e. `MyLib.loadMyConfig`).

There is also an Intellisense issue regardless of whether the workaround is used: If you attempt to dot into `MyLib` (i.e. type `MyLib.`), Intellisense will show the same error reported above.

I have put together a repro here: https://github.com/ScottShingler/JsonProviderEmbeddedRepro

I should also mention that I'm experiencing the same compilation error in another project, but the workaround of not opening the library module is not solving it. Unfortunately I am unable to share that code since it's proprietary, but I expect the root cause of this issue should be related.

ArtemyB:
@ScottShingler just had the same issue too, but with `XmlProvider`. As I understand the problem is in the format of embedded resource names used in .NET Standard/.NET Core. Try to specify the resource name as follows: `EmbeddedResource = "MyLib, MyLib.sample.json"`. Hope it works for you as it worked for me. :)

P.S.: [This SO answer](https://stackoverflow.com/questions/38762368/embedded-resource-in-net-core-libraries) helped me.

fischgeek:
I'm experiencing the same issue with the JsonProvider and the workaround doesn't seem to work for me.

**Lib**

    namespace TrelloConnectFSharp
    open FSharp.Data
    module Trello = 
[...]

PR #1334, opened 2020-11-20
Update JsonProvider.fsx

psfinaki:
I was struggling with that and things were not working until I found [this comment](https://github.com/fsharp/FSharp.Data/issues/1191#issuecomment-609921277) noting that point about project file. I think it would be useful for others as well.

### DIFF
--- a/docs/content/library/JsonProvider.fsx
+++ b/docs/content/library/JsonProvider.fsx
@@ -295,7 +295,8 @@ access the sample JSON. This works fine when the sample is specified inline, but
 the sample is specified as a local file (unless you distribute the samples with your library).
 
 For this reason, the JSON provider lets you specify samples as embedded resources using the 
-static parameter `EmbeddedResource`. If you are building a library `MyLib.dll`, you can write:
+static parameter `EmbeddedResource` (don't forget then to include the file as EmbeddedResource in the
+project file). If you are building a library `MyLib.dll`, you can write:
 
 *)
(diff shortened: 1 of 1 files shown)


nhirschey:
1. In dotnet core, does "MyLib" also need to be added to the embedded embedded resource string? Such as in

```fsharp
type WB = JsonProvider<"../data/WorldBank.json", 
 EmbeddedResource="MyLib, MyLib.worldbank.json"
```

  I'm looking at https://github.com/fsharp/FSharp.Data/blob/08f351820f8b44c21ac6fb5afb30be0236332d0d/tests/FSharp.Data.Tests/JsonProvider.fs#L19

2. It would also be nice to show explicitly how to include the embedded resource in the project file.

psfinaki:
@nhirschey,
1. Well my case was dotnet core, so yes
2. Added a link to MSDN, there is also a screenshot with options, I guess it should be sufficient

nhirschey:
Yeah, my point is to make dotnet core the default, so change line 303 to

```fsharp
  EmbeddedResource="MyLib, MyLib.worldbank.json">
```

nhirschey:
I also ran into this last week with FSharp.Data.Toolbox, and knowing that in dotnet core you need MyLib at the front would have helped. Just trying to help the next poor souls so they don't have our issues ;)

psfinaki:
Oh that thing, yeah sure, added the path there :)

PR #1344, opened 2021-01-13
add ResolutionFolder=__SOURCE_DIRECTORY__ to docs

dsyme:
Because of https://github.com/fsprojects/FSharp.Data/issues/1340 and https://github.com/fsprojects/FSharp.Data/issues/1306 we should just ask people to add `ResolutionFolder=__SOURCE_DIRECTORY__` to all examples that use relative path names.


nhirschey:
Can we deploy the docs with this fix? It's not showing up for me at http://fsprojects.github.io/FSharp.Data/library/CsvProvider.html

dsyme:
> Can we deploy the docs with this fix? It's not showing up for me at http://fsprojects.github.io/FSharp.Data/library/CsvProvider.html

This is now showing (though not on the animated gif which uses a direct path)

PR #1348, opened 2021-01-15
fix links in docs/tools/generate.fsx

HKGx:
Recently FSharp.Data was moved into fsprojects organization breaking some links.


sergey-tihon:
Thank you @HKGx !

PR #1357, opened 2021-02-21
Improved missing values documentation for the csv provider

nhirschey:
1. Improved the docs. The additional strings that I note as default missing values come from https://github.com/fsprojects/FSharp.Data/blob/4697bfdf4e5ee9d761695f8432a944ae92947143/src/CommonRuntime/TextConversions.fs#L43
2. Improved the API help text.


cartermp (approved):
Thanks!

PR #1360, opened 2021-03-04
Build and test against .NET 5

cartermp:
* Updates the build to .NET 5
* Updates tests to .NET 5
* Simplify the type provider project files by using F# compiler properties
* Use paket throughout the repo
* Remove linqpad samples from the package

After this change, NPE shows the following:

image

The red checks are the same as with the current version of the package if you build it locally. Symbols aren't actually missing, they're just a separate snupkg that is also built.


cartermp:
All tests pass locally now, but something is still weird in CI.

cartermp:
This should not be merged yet. I think there is an issue with the FSharp.Core dependency in paket. It's not writing the correct version dependency when packing. Ugh.

dsyme (approved):

dsyme:
Thank you!

cartermp:
Now to clean a few more things up and publish a preview!

PR #1365, opened 2021-04-14
Fix internet cache invalidation

laenas:
Bug was spotted and reported on Slack today - as it stands, nothing is ever invalidated because we pretty much eternally live in negative date-time space, which will always be less than a positive TimeSpan.


cartermp (approved):
Thanks!

PR #1370, opened 2021-04-15
update docs

dsyme:
Fixes the issues in https://github.com/fsprojects/FSharp.Data/issues/1368

Also move some docs from non-standard markdown to standard XML 

Note the fsdocs sidebar has moved the left by default.  I'm also thinking we should change the default fsdocs template to position it more like [DiffSharp](https://diffsharp.github.io/) does than all the way to the left as the default is now doing. (I'd like to keep using the FSharp.Formatting default template so we can keep making improvements there)

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -2,4 +2,4 @@
 
 dotnet tool restore
 dotnet paket restore
-dotnet fake build -t Build
+dotnet fake build  %*
--- a/build.fsx
+++ b/build.fsx
@@ -32,9 +32,9 @@ let project = "FSharp.Data"
 let authors = "Tomas Petricek;Gustavo Guerra;Colin Bull;fsprojects contributors"
 let summary = "Library of F# type providers and data access tools"
 let description = """
-  The FSharp.Data library (FSharp.Data.dll) contains type providers and utilities to access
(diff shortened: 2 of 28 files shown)


nhirschey:
Strong second of a sidebar along the lines of the diffsharp docs template. 

I’ve been meaning to raise an issue, but if you’re on a mobile browser then clicking the sidebar on the current template fills the browser window with the sidebar, and that sidebar menu that pops out is not scrollable. So it’s impossible (at least on iOS) to scroll to and click on links at the bottom of the side bar. You’re restricted to the first few items on the sidebar. In below screenshot, I can’t access anything below XML Provider.

Diffsharp has sidebar links at the top on mobile, which makes it scrollable. 
6D28F68B-A623-4FDA-8211-F90C37CABBB4

dsyme:
I think the problem is that the intention is that the nav bar is scrollable and fixed-location (which works on desktop).  And the DiffSharp one has no hamburger on mobile

Anyway could you file an issue in FSharp.Formatting? Thanks

dsyme:
> Anyway could you file an issue in FSharp.Formatting? Thanks

(The same issue applied to the FSharp.Formatting site AFAIK)

PR #1377, opened 2021-04-26
set sourcelink to be a private asset so that it doesn't get added to package dependencies

### LINKED ISSUE #1376: Question about new dependencies in 4.1.0
This version has some new dependencies, but my changed project file just reflects the new version of 4.1.0 for FSharp.Data. So, I don't need to worry about these other packages? I'm targeting .NET Framework 4.7.2.

image

I googled one of these and saw something about support for debugging Nuget binaries, maybe there is a blog post somewhere that succinctly explains all this?

baronfel:
Fixes #1376 to remove the errant sourcelink dependency.

After this change, the nuspec contains the following dependency groups:

```xml
<dependencies>
  <group targetFramework=".NETStandard2.0">
    <dependency id="FSharp.Core" version="4.7.2" exclude="Build,Analyzers" />
  </group>
</dependencies>
```

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -12,7 +12,7 @@ github fsprojects/FSharp.TypeProviders.SDK src/ProvidedTypes.fs
 github fsprojects/FSharp.TypeProviders.SDK tests/ProvidedTypesTesting.fs
 
 nuget FSharp.Core >= 4.7.2
-nuget Microsoft.SourceLink.GitHub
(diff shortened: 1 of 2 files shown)


baronfel:
Test errors are from the worldbank data, but don't represent a problem with the package.

cartermp (approved):
Thanks @baronfel

Issue #1235, opened 2018-12-07
Please sign the assembly


astrowalker:
For the record I found related issue: https://github.com/fsharp/FSharp.Data/issues/990

The problem is the assemblies are not signed so when they propagate you end up with non-signed assembly at the end. When it is fine, it is fine, but when signed is required you are toasted.

The point is, please sign assemblies.

tearvisus:
I agree. Strong naming has few drawbacks nowadays. NuGet automatically generates necessary assembly binding redirects.
[The .NET open-source library guide](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries) recommends considering strong naming as well.

jmecosta:
i think this is really needed, workarounds like strongnamer nuget do not work becasue of design assemblies. 

1>C:\Program Files\dotnet\sdk\3.0.100-preview5-011568\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(157,5): message NETSDK1057: You are using a preview version of .NET Core. See: https://aka.ms/dotnet-core-preview
1>FSC : error FS3031: The type provider 'D:\prod\structures\ObjDrop\Work\Debug\x64\FarmiCaseHelpers\SignedAssemblies\FSharp.Data.dll' reported an error: Assembly attribute 'TypeProviderAssemblyAttribute' refers to a designer assembly 'FSharp.Data.DesignTime' which cannot be loaded or doesn't exist. Could not load file or assembly 'file:///D:\prod\structures\ObjDrop\Work\Debug\x64\FarmiCaseHelpers\SignedAssemblies\FSharp.Data.DesignTime.dll' or one of its dependencies. The system cannot find the file specified.
1>FSC : error FS3005: Referenced assembly 'D:\prod\structures\ObjDrop\Work\Debug\x64\FarmiCaseHelpers\SignedAssemblies\FSharp.Data.dll' has assembly level attribute 'Microsoft.FSharp.Core.CompilerServices.TypeProviderAssemblyAttribute' but no public type provider classes were found
1>Done building project "FarmiCaseHelpers.fsproj" -- FAILED.
2>------ Build started: Project: FarmiAdapterUtils, Configuration: Debug Any CPU ------
2>FarmiAdapterUtils -> D:\prod\structures\BuildDrop\Work\bin_x64\Test\TestAdapters\FarmiAdapterUtils.dll

[...]

Issue #990, opened 2016-08-26
Getting runtime error: Could not load file or assembly 'FSharp.Data... A strongly-named assembly is required.


nikoudel:
This probably happens because FSharp.Data.dll is not signed. Is there some reason why it's not signed?

`System.IO.FileLoadException was unhandled by user code
  FileName=FSharp.Data, Version=2.3.2.0, Culture=neutral, PublicKeyToken=null
  FusionLog=""
  HResult=-2146234300
  Message=Could not load file or assembly 'FSharp.Data, Version=2.3.2.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)`

nikoudel:
I've ended up signing FSharp.Data.dll [manually](http://stackoverflow.com/questions/1379954/how-i-do-a-sign-an-assembly-that-has-already-been-built-into-a-dll-specifically) but now I have to do it every time FSharp.Data updates. Would be nice to have it readily signed.

nikoudel:
Can't reproduce the issue with F# 4 any more.

PR #1386, opened 2021-07-26
Sign assemblies with strong name key file

JosephFerano:
This PR addresses #990 and #1235


JosephFerano:
I attempted to use public signing, but ran into a similar issue to this:

https://github.com/dotnet/machinelearning/pull/600#issuecomment-408869433

If you would prefer having the snk file somewhere else, let me know and I'll gladly move it to it's appropriate location.

dsyme:
See comment here, https://github.com/fsprojects/FSharp.Data/issues/1235#issuecomment-887109417

dsyme (approved):

JosephFerano:
Thank you for merging. The reason for the strong naming is that our .netstandard project was recently integrated into one of our products. It's a plugin for an older system that only supports .Net framework and assemblies have to be signed.

PR #1389, opened 2021-08-10
Removing snk file from fsproj references

JosephFerano:
I accidentally included the snk file in the fsproj, which caused nuget to include the reference when building the nupkg and subsequently deploying it along with the DLL. This led to the snk file being installed in all projects that referenced FSharp.Data. 

My apologies.

### DIFF
--- a/src/FSharp.Data.DesignTime/FSharp.Data.DesignTime.fsproj
+++ b/src/FSharp.Data.DesignTime/FSharp.Data.DesignTime.fsproj
@@ -61,9 +61,6 @@
     <Compile Include="..\Html\HtmlGenerator.fs" />
     <Compile Include="..\Html\HtmlProvider.fs" />
     <Compile Include="..\AssemblyInfo.DesignTime.fs" />
-    <None Include="..\keyfile.snk">
-      <Link>keyfile.snk</Link>
-    </None>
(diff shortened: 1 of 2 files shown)


ReedCopsey:
@cartermp This, unfortunately, "pollutes" any project with the .snk file if it uses the 4.2.0 nuget package.

cartermp (approved):
Thanks! No worries :)

ReedCopsey:
Thanks @cartermp - Any way we can help to get an updated 4.2.1 nuget out? Would be good to not leave this causing issues.

dsyme:
I've updateded RELEASE_NOTES.md, which, if all is going correctly in the auto-push, should build and push the nuget

dsyme:
https://github.com/fsprojects/FSharp.Data/commit/d2b1bafb9fda785da369b9fb28b9a382e25d8b61

cartermp:
It will not unless the assembly info changes as well AFAIK

dsyme:
> It will not unless the assembly info changes as well AFAIK

Pushed changes with 4.2.2

ReedCopsey:
@dsyme and @cartermp Thank you. New package doesn't have the snk file now.

PR #1393, opened 2021-09-15
@tomakita's ampersand parsing fix

njlr:
See https://github.com/fsprojects/FSharp.Data/issues/1264


cartermp (approved):
And thanks to cleaning up the CI and build, this puppy runs.

PR #1395, opened 2021-10-11
multipart/form-data must use CRLF line breaks

### LINKED ISSUE #1313: Multi-Part HTTP Request Body on Linux
Having an issue with multi-part bodies resulting in a 400 response on Linux. Was able to craft an `HttpWebRequest` that worked, so narrowed it down to the library.

I suspect the issue is the use of `Environment.NewLine` used to create `headerpart` in `writeMultipart`.

If anyone can confirm my expectation that this should be a literal "\r\n" similar to `newlineStream` I could submit a patch.

kalekseev:
Fixes #1313
`Environment.NewLine` equals `\n` on unix like systems that leads to invalid payload with LF line breaks:
```
--define a custom boundary here<LF>
Content-Disposition: form-data; name="images"; filename="image0.png"<LF>
Content-Type: image/png<CRLF>
...
```

From https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html 

>  The boundary must be followed immediately either by another CRLF and the header fields for the next part, or by two CRLFs, in which case there are no header fields for the next part (and it is therefore assumed to be of Content-Type text/plain).

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1359,7 +1359,7 @@ module private HttpHelpers =
                 [ prefixedBoundary
                   HttpRequestHeaders.ContentDisposition("form-data", Some formField, Some fileName) |> printHeader
                   HttpRequestHeaders.ContentType contentType |> printHeader ]
-                |> String.concat Environment.NewLine
+                |> String.concat "\r\n"
             let headerStream =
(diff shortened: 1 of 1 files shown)


cartermp:
I think this looks sensible. @baronfel, since you added the feature, any concerns?

PR #1397, opened 2021-10-14
Correction to positioning of boundary delimiter with multiple multipart/form-data files

ordinaryorange:
newline required before writing next boundary and header of subsequent files, otherwise server will only decode first file.

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1367,7 +1367,8 @@ module private HttpHelpers =
                 [ headerStream
                   newlineStream()
                   newlineStream()
-                  fileStream ]
+                  fileStream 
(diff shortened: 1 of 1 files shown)


dsyme:
@ordinaryorange I shouldn't have merged this without a test and a tracking issue

Could you add both please?
Thanks

ordinaryorange:
Done.
I've had to make the `HttpUtilities` module in [http.fs](https://github.com/fsprojects/FSharp.Data/blob/9bc33bb5db9549b02fd3686fbe0a0cda075a9a7c/src/Net/Http.fs#L1239) `internal` for the test. I assume this is ok ?

PR #1403, opened 2021-11-09
Update docs

dsyme:


cartermp (approved):
Giving this an approval since I agree with the changes broadly. Assuming it's tweaked to your liking and CI passes, 🚢 🇮🇹

dsyme:
Note that a typical "dotnet paket update" caused lots of package bumps which caused lots of hassle.  So I pinned loads of packages to prevent the paket update changing much.

PR #1408, opened 2021-11-17
Add plural for "source"

### LINKED ISSUE #1318: Odd naming in JsonProvider
The following provided type will _not_ have a type called Config.Sources but instead have a type called Config.Sourcis. 

`type Config = JsonProvider<"""{"sources": [{"a":"b"}]}""">`

on the other hand 

`type Config = JsonProvider<"""{"sources": {"a":"b"}}""">`

does have a type called Config.Sources

jimfoye:
This fixes #1318.


cartermp (approved):
Thanks!

PR #1414, opened 2021-11-19
HtmlState: restore FormattedMode if still in pre after char ref

stroborobo:
At the same time the assumption that the code element is formatted is not correct. They are often used at the same time though and with that comes another issue: keeping FormattedMode in elements nested into pre.

I'm not sure how to approach this, I've never worked with refs like here before. Would it be ok to drop the FormattedMode and add an IsFormatted: bool ref instead? And update that when entering/exiting pre elements.

### DIFF
--- a/src/Html/HtmlParser.fs
+++ b/src/Html/HtmlParser.fs
@@ -334,7 +334,10 @@ module internal HtmlParser =
             x.Tokens := result :: !x.Tokens
 
         member x.IsFormattedTag
-            with get() = x.CurrentTagName().ToLower() = "pre"
+            with get() =
+                match x.CurrentTagName().ToLower() with
+                | "pre" | "code" -> true
+                | _ -> false
 
         member x.IsScriptTag
             with get() =
--- a/src/Html/HtmlParser.fs
+++ b/src/Html/HtmlParser.fs
@@ -334,14 +334,11 @@ module internal HtmlParser =
             x.Tokens := result :: !x.Tokens
 
         member x.IsFormattedTag
-            with get() =
-               match x.CurrentTagName() with
-               | "pre" | "code" -> true
-               | _ -> false
+            with get() = x.CurrentTagName().ToLower() = "pre"
 
         member x.IsScriptTag
             with get() =
-               match x.CurrentTagName().Trim().ToLower() with
+               match x.CurrentTagName().ToLower() with
                | "script" | "style" -> true
                | _ -> false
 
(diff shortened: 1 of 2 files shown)


tests/FSharp.Data.Tests/HtmlParser.fs
```diff
@@ -805,15 +805,16 @@ let ``Can parse pre blocks``() =
     result |> should equal [ "\r\n        This code should be indented and\r\n        have line feeds in it" ]
 
 [<Test>]
-let ``Can parse code blocks``() =
-    let html = "<code>\r\n        let f a b = a * b\r\n        f 5 6 |> should equal 30</code>"
+let ``Can parse pre blocks with char refs``() =
```
cartermp, line 808: This removes the code blocks test - could you add that test back as well?

stroborobo: Sure, it would test for invalid behavior though since code is an inline element. Should I restore the behavior for code anyways and open another issue for elements nested in pre?

stroborobo:
Hey @cartermp, since you didn't say anything I implemented my earlier suggestion with the bool ref instead. My assumption was not correct in the original implementation anyways, so it didn't really work well, test was flawed also :/

But this seems to work, at least for the fixed up test and my use case of glueing together different parts of fsharp.github.io/fsharp-core-docs

src/Html/HtmlParser.fs
```diff
@@ -275,13 +273,15 @@ module internal HtmlParser =
         { Attributes : (CharList * CharList) list ref
           CurrentTag : CharList ref
           Content : CharList ref
+          HasFormattedParent: bool ref
```
cartermp, line 276: An alternative is that this can be a `mutable HasFormattedParent` to avoid the `ref`, but I don't think that's terribly important.

src/Html/HtmlParser.fs
```diff
@@ -355,9 +355,15 @@ module internal HtmlParser =
                     else TagEnd(name)
                 else Tag(false, name, x.GetAttributes())
 
+            // pre is the only default formatted tag, nested pres are not
+            // allowed in the spec.
+            if x.IsFormattedTag then
+                x.HasFormattedParent := not isEnd
+            else
+                x.HasFormattedParent := !x.HasFormattedParent || x.IsFormattedTag
```
cartermp, line 363: as an example if this is `mutable`-able, this could just be

`x.HasFormattedParent <- x.HasFormattedParent || x.IsFormattedTag`

The codebase is fairly old :)

cartermp:
Weird that this failure is only on Windows: `Drops whitespace outside pre`

stroborobo:
Oh right, the formatter inserts Environment.NewLine, totally forgot about it. I'll change the test to expect those as well.

src/Html/HtmlParser.fs
stroborobo: Hm, I just meant to be consistent, should I update it?

Issue #1398, opened 2021-10-28
Multipart form files only accept seekable streams


ordinaryorange:
I've a situation where I need to pass in a [PipeReader](https://docs.microsoft.com/en-us/dotnet/api/system.io.pipelines.pipereader?view=dotnet-plat-ext-5.0) stream for use a `MiltiPartItem` form files for a `POST` request.

However I notice the current [writeMultipart](https://github.com/fsprojects/FSharp.Data/blob/9665fc39e22b492634881208f3f98a3107469d4e/src/Net/Http.fs#L1351) and [CombinedStream](https://github.com/fsprojects/FSharp.Data/blob/9665fc39e22b492634881208f3f98a3107469d4e/src/Net/Http.fs#L1310) implementations lean on the `Stream.Length` property which is where things fail.

The `Length` property assumes provided streams are seekable, which `PipeReader` (and others) are not.

In my testing I've removed all the `Stream.Lengh` references, (which also means the `CombinedStream` cant calculate it's length) and removed the assignment of length during [request](https://github.com/fsprojects/FSharp.Data/blob/9665fc39e22b492634881208f3f98a3107469d4e/src/Net/Http.fs#L1395) building. 

Requests are still issued to and accepted by the server, so it initially seems ok to permit non seekable streams

Thoughts

- This is using the current `WebRequest` implementation, and I've no idea if this will work in `HttpClient` #1392 
- I'm unsure if it is a requirement for a `POST` request to have a Length.
- Are non seekable streams for multipart items even a valid idea for inclusion
- Anything else I've overlooked

Thoughts please, as to if this would be a valid inclusion ?

ordinaryorange:
I've done an implementation and tests [here](https://github.com/ordinaryorange/FSharp.Data/commit/f644892b7d67aaf7897af1c2bdae069a33f44afe)

Main points
- I've added all the checks for non seekable streams where the content length cannot be determined up front.

- I've also refactored the `CombinedStream` type to better report its `Length` and `CanSeek` implementations to be closer to the expected for `Stream` behaviour (even though `CombindStream` is totally internal)

- And refactored the calculation of content length to be as light as possible with the `IEnumerator` on the incoming seq. (This was something I needed specifically for my passing of a `BlockingCollection` but should work with all others)

All tests pass, and implementation is working well in my app

Can someone review and let me know if ok for a PR ?

cartermp:
[...]

PR #1415, opened 2021-11-23
Implementation and tests - non seekable multipart form files

ordinaryorange:
Refer #1398

### DIFF
--- a/src/Net/Http.fs
+++ b/src/Net/Http.fs
@@ -1313,11 +1313,11 @@ module internal HttpHelpers =
             let mutable streams = streams |> Seq.cache
 
             let rec readFromStream buffer offset count =
-                if Seq.isEmpty streams
-                then 0
-                else
-                    let stream = Seq.head streams
-                    let read = stream.Read(buffer, offset, min count (int stream.Length))
+                match streams |> Seq.tryHead with
+                | None -> 0
+                | Some stream ->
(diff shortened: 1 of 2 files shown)


src/Net/Http.fs
```diff
@@ -1327,9 +1327,9 @@ module internal HttpHelpers =
                     else read
 
             override x.CanRead = true
-            override x.CanSeek = false
+            override x.CanSeek = match length with | None -> false | Some _ -> true
             override x.CanWrite = false
-            override x.Length with get () = length
+            override x.Length with get () = length |> Option.defaultWith (fun () -> NotSupportedException() |> raise)
```
cartermp, line 1332: What do you think about a more descriptive message like in `Stream`? https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.length?view=net-6.0#exceptions

ordinaryorange: If flipped on this a couple of times. But now rethinking, I reckon a simple `failwith "One or more of the encompassed streams are not seekable and the length cannot be determined"`. We can probably do away with the `NotSupportedException` type as the `CombinedStream` is only internal to `FSharp.data`

cartermp (approved):
Thank you! Will cut a release likely beginning of next week.

cartermp:
Should be live in NuGet promptly

Issue #1420, opened 2022-01-01
HTTP - No need extra newline before the endBoundaryStream in writeMultipart


kphuanghk:
Regarding #1397 , after introduced newline in MultipartItem stream construction. It should no longer need to have newline before the `endBoundaryStream` . This will cause extra 2 bytes "/r/n" to the last multipart item. 

Original:
```fsharp
        let wholePayload = Seq.append segments [newlineStream(); endBoundaryStream; ]
```
Shall be:
```fsharp
        /// per spec, close-delimiter := "--" boundary "--" CRLF ; no need extra newline
        let wholePayload = Seq.append segments [ endBoundaryStream; ]
```

Here's the skeleton program I used to test the two extra bytes issue:
[TestProgram.fs.MD](https://github.com/fsprojects/FSharp.Data/files/7797880/TestProgram.fs.MD)

 For those affected by this issue, just need to remove extra two bytes for the affected files.

cartermp:
Addressed in https://github.com/fsprojects/FSharp.Data/pull/1421

PR #1421, opened 2022-01-01
Proposed fix to #1420 on extra newline before endBoundaryStream

kphuanghk:
As reported in #1420 , two extra bytes of newline is appended to the last Multipart Item, this PR aims to fix it.

Since extra newline removed, two existing test cases shall be revised to reflect this change:

- ``correct multipart content format``
- ``Seekable streams create Seekable CombinedStream``

I have tested several scenarios on number of files,  format, size etc. upon the built FSharp.Data.dll and looks good.


cartermp (approved):
Thanks!

PR #1422, opened 2022-01-01
Build on .NET 6

cartermp:


cartermp:
FAKE isn't stable enough to do this yet.

baronfel:
If you want to go ahead anyway, there are example paths forward here: https://github.com/fsprojects/FAKE/issues/2618#issuecomment-998822202

PR #1424, opened 2022-01-08
JsonProvider.Load(value: JsonValue)

### LINKED ISSUE #1309: Support loading data from JsonValue for JsonProvider.
JsonProvider supports loading data from uri, stream, and stream reader, but not `JsonValue` which is inconvenient.

Now I had to use it like this:
```FSharp
use ms = new MemoryStream()
use w = new StreamWriter(ms, leaveOpen=false)
value.WriteTo(w, JsonSaveOptions.DisableFormatting)
w.Flush()
ms.Seek(0L, SeekOrigin.Begin) |> ignore
let record = DayCloseData.Load(ms)
```
Instead of just `let record = = DayCloseData.Load(value)`

---
If you a curious why I need this, I can explain. <br/>
Some services return such payload:
```js
{
    "history": {
        "2020-01-10": {
            ...
        },
        "2020-01-11": {
            ...
        }
    }
}
```

gbtb:
Added overload of Load method on JsonProvider to support combining JsonProviders without need for additional conversion to and from string. This should fix #1309

### DIFF
--- a/tests/FSharp.Data.DesignTime.Tests/SignatureTests.fs
+++ b/tests/FSharp.Data.DesignTime.Tests/SignatureTests.fs
@@ -28,7 +28,13 @@ let expectedDirectory = sourceDirectory ++ "expected"
 
 let resolutionFolder = sourceDirectory ++ ".." ++ "FSharp.Data.Tests" ++ "Data"
 let assemblyName = "FSharp.Data.dll"
-let netstandard2RuntimeAssembly = sourceDirectory ++ ".." ++ ".." ++ "src" ++ "FSharp.Data" ++ "bin" ++ "Release" ++ "netstandard2.0" ++ assemblyName
+let netstandard2RuntimeAssembly = sourceDirectory ++ ".." ++ ".." ++ "src" ++ "FSharp.Data" ++ "bin" ++
+                                #if DEBUG
+                                    "Debug"
+                                #else
+                                    "Release"
+                                #endif
+                                ++ "netstandard2.0" ++ assemblyName
 
 let getRuntimeRefs platform = TypeProviderInstantiation.GetRuntimeAssemblyRefs platform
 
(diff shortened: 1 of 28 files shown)


tests/FSharp.Data.DesignTime.Tests/SignatureTests.fs
```diff
@@ -28,7 +28,13 @@ let expectedDirectory = sourceDirectory ++ "expected"
 
 let resolutionFolder = sourceDirectory ++ ".." ++ "FSharp.Data.Tests" ++ "Data"
 let assemblyName = "FSharp.Data.dll"
-let netstandard2RuntimeAssembly = sourceDirectory ++ ".." ++ ".." ++ "src" ++ "FSharp.Data" ++ "bin" ++ "Release" ++ "netstandard2.0" ++ assemblyName
+let netstandard2RuntimeAssembly = sourceDirectory ++ ".." ++ ".." ++ "src" ++ "FSharp.Data" ++ "bin" ++
+                                #if DEBUG
+                                    "Debug"
+                                #else
+                                    "Release"
+                                #endif
```
cartermp, line 36: Why this change?

cartermp (commented):
Looks good overall, curious about the test change

tests/FSharp.Data.DesignTime.Tests/SignatureTests.fs
gbtb: Maybe it's just my workflow, but I rarely build projects in a Release configuration while developing. So I've ran tests in debug mode and was a bit puzzled when all of them failed at once, because of missing Release-build dll

cartermp: Cool!. I think this is fine. I had to make a change when building/testing for .NET 5 to run tests in release, since at the time the compiler could stack overflow. That may no longer be the case (there have been a years' worth of compiler improvements since then). This is probably a better way to do it.

PR #1433, opened 2022-03-29
Typo fix, the parameter is plural ("Separators")

nhirschey:
See https://fsprojects.github.io/FSharp.Data/reference/fsharp-data-csvprovider.html#Separators


cartermp (approved):
Thanks!

Issue #1292, opened 2019-10-23
Can't get inner text of a script


LiteracyFanatic:
Is there some reason getting the inner text of a script element is not allowed? I'm working on a webscraping project where I need need to obtain a cookie by evaluating some javascript from one of the pages. I'm shelling out to Node.js to do that, but the rest of the HTML parsing and logic is all in F#.  I can work around this by manually stripping the script tags myself, but it'd be much more convenient to just call `HtmlNode.innerText`. It looks like this restriction was introduced in commit a715fbc8ac1b40ebe5b4aa9bb1bbc695e38d4b6b, but I can't see any explanation as to why.

pedersenjens:
Yes, why is it that that inner text is not available for script tags and apparently also not style tags? Is there a good reason for this change?

tnishimura:
I also find this an odd bug, any chance of reverting the changes?

dsyme:
I don't know of any intrinsic reason - we can't revert a change from that long ago but please do submit a PR to address this if you can

PR #1435, opened 2022-04-04
Remove restrictions for getting innerText

LiteracyFanatic:
This pull request addresses #1292. It seems like the current version of `innerTextExcluding'` was tightly coupled to some of the implementation details of the Html type provider. In order to avoid breaking the `Validate signature didn't change` tests, I moved the existing logic to private functions inside `HtmlRuntime.fs`. Then I added a new simpler implementation to `HtmlOperations.fs` which doesn't modify the exclusion list or filter on the `aria-hidden` attribute.


dsyme:
Fabulous!

PR #1441, opened 2022-06-01
Format all code

nojaf:
Hello, would you be interested in using Fantomas for this community project?


baronfel:
even if there is a reason for sticking to `net5.0` target frameworks, the project itself can and should move to 6.0 SDKs. You can target older TFMs from newer SDKs, you just also get the benefit of any fixes/feature enhancements that came with the newer SDK.

baronfel:
Yes, with a caveat - the PR/merge workflows use the `setup-dotnet` action to [ensure the appropriate version of the .NET SDK is installed](https://github.com/fsprojects/FSharp.Data/blob/main/.github/workflows/pull-requests.yml#L14-L17). If you change the value in global.json, I would also change these usages over to install multiple .NET SDKs to ensure that tests still run:

```yaml
- name: Setup dotnet
  uses: actions/setup-dotnet@v2
  with:
    dotnet-version: | 
      5.0.x
      6.0.x
```

dsyme:
Reformatting may be needed after pull

Regarding .NET version - we can update it - it might require an update to paket and fake?

src/CommonProviderImplementation/Helpers.fs
```diff
@@ -202,169 +233,210 @@ module internal ProviderHelpers =
...
+        (tp: DisposableTypeProviderForNamespaces)
+        (cfg: TypeProviderConfig)
+        encodingStr
+        resolutionFolder
+        resource
+        fullTypeName
+        maxNumberOfRows
+        =
+
+        using (logTime "LoadingTextToBeParsed" valueToBeParsedOrItsUri)
```
dsyme, line 249: I will pre-change all explicit `using` to `use _holder = ...` to prevent this indentation happening.  Then we reformat

cartermp:
Updating to net6 is possible, it will just be annoying. IIRC this requires FAKE to move to be project-based

src/Csv/CsvGenerator.fs
```diff
@@ -15,79 +15,121 @@ open ProviderImplementation.QuotationBuilder
...
+            | [ field ] -> field.TypeForTuple
+            | _ -> FSharpType.MakeTupleType([| for field in fields -> field.TypeForTuple |])
+
+        let rowType =
+            ProvidedTypeDefinition("Row", Some rowErasedType, hideObjectMethods = true, nonNullable = true)
+
+        let ctor =
+            ProvidedConstructor(
+                [ for field in fields -> field.ProvidedParameter ],
+                invokeCode =
```
dsyme, line 73: These `invokeCode` expand out a fair bit, I will pre-adjust them to be inner function bindings

src/Xml/XmlRuntime.fs
```diff
@@ -58,54 +80,84 @@ open System.Xml.Linq
...
+        use reader = reader
+        let text = reader.ReadToEnd()
+
+        let element =
+            XDocument
+                .Parse(
+                    text,
+                    LoadOptions.PreserveWhitespace
+                )
+                .Root
```
dsyme, line 135: This is strangely expanded.  Is line width 80?  I think it could be set to 120.

dsyme:
@nojaf Could you reformat after https://github.com/fsprojects/FSharp.Data/pull/1442 please?

@cartermp We should likely just cut FAKE out of this project

dsyme:
@nojaf I'm doing the update to .NET 6 separately here: https://github.com/fsprojects/FSharp.Data/pull/1443

dsyme:
I'll make a couple more tweaks to .edtiorconfig in a separate PR

dsyme:
@nojaf Thank you sooooooo much for pushing this along.  It makes the repository so much easier to maintain and contribute to

nojaf:
You are most welcome. The more projects are formatted, the more exposure for Fantomas 😸.

PR #1444, opened 2022-06-02
formatting: wider lists

dsyme:
Tweak the formatting settings

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,7 @@ insert_final_newline = true
 trim_trailing_whitespace = false
 
 [*.fs]
+fsharp_max_array_or_list_width=80
 fsharp_max_array_or_list_number_of_items=5
 fsharp_max_if_then_else_short_width=60
 fsharp_max_dot_get_expression_width=80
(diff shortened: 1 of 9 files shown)


.editorconfig
```diff
@@ -13,6 +13,7 @@ insert_final_newline = true
 trim_trailing_whitespace = false
 
 [*.fs]
+fsharp_max_array_or_list_width=80
```
dsyme, line 16: @nojaf It does feel to me that these defaults can be wider for F# code, likewise the other settings here

.editorconfig
```diff
@@ -13,6 +13,7 @@ insert_final_newline = true
...
 fsharp_max_array_or_list_number_of_items=5
```
nojaf, line 17: `fsharp_max_array_or_list_number_of_items` does not do anything on its own.
You need to add `fsharp_array_or_list_multiline_formatter=number_of_items` for this to do anything. And please don't, I regress this setting exists in the first place.

.editorconfig
nojaf: I've opened https://github.com/fsprojects/fantomas/issues/2288 to discuss this.

PR #1450, opened 2022-06-12
Some medium refactoring and little fixes

mlaily:
This is a series of commits I did while working on #1447 and #1448 that were not strictly necessary for the aforementioned PRs to work, but I think it might be nice to have anyway.

Changes:
- Fix an out of bound exception I saw while debugging the providers. It might not happen when the providers work, but now it will never crash and is better documented.
- We have a `PrimitiveInferedProperty` to represent a type and a name, but there are several cases where we just want a type, so I added a `PrimitiveInferedValue` to do just that. The core or #1447 was initially based on this change, but in the end I did it another way.
- As seen in #1448, there is a test that ensures an xsd is seen as invalid. Unfortunately, this test outputs an error message in the console that is very confusing because the test works. I suppressed this message.

If you feel like some commits are undesirable, I can remove them...  
Oh, and this should merge cleanly with #1447 BTW. Unless the auto-formatting messed something after I checked.


cartermp (approved):
Thanks!

PR #1451, opened 2022-06-20
Fix http tests

mlaily:
The CI for #1447 already failed two times, so let's see if this PR improves the situation...

I initially thought the problem was simply that using an external (internet) endpoint for the tests was unreliable, but the tests were still failing a bit randomly after replacing the remote endpoints with a local http test server.

After investigation, I believe there is a race condition between the http stack timeout and the `Async.StartChild` timeout. (in `getResponse` in `Http.fs`)  
If the http stack timeout is triggered first, the inner exception is null. Since the tests were asserting that the inner exception should be a `TimeoutException`, they were failing in this case.

In all cases, the `WebException` status should indicate a timeout, so I changed the assertion accordingly...

### DIFF
--- a/tests/FSharp.Data.Tests/Http.fs
+++ b/tests/FSharp.Data.Tests/Http.fs
@@ -8,17 +8,78 @@ open System.Net
 open FSharp.Data
 open FSharp.Data.HttpRequestHeaders
 open System.Text
+open System.Threading.Tasks
+open Microsoft.AspNetCore.Builder
+open Microsoft.AspNetCore.Http
+open System.Net.NetworkInformation
+
+type ITestHttpServer =
+    inherit IDisposable
+    abstract member BaseAddress: string
+    abstract member WorkerTask: Task
+
+let startHttpLocalServer() =
+    let app = WebApplication.CreateBuilder().Build()
+    app.Map("/{status}", (fun (ctx: HttpContext) ->
+        async {
+            match ctx.Request.RouteValues.TryGetValue("status") with
+            | true, (:? string as status) ->
+                let status = status |> int
+
+                match ctx.Request.Query.TryGetValue("sleep") with
+                | true, values when values.Count = 1 ->
+                    let value = values[0] |> int
+                    do! Async.Sleep value
+                | _ -> ()
+
+                if ctx.Request.Body <> null then
+                    let buffer = Array.create 8192 (byte 0)
+                    let mutable read = -1
+                    while read <> 0 do
+                        let! x = ctx.Request.Body.ReadAsync(buffer, 0, 8192) |> Async.AwaitTask
+                        read <- x
+
+                Results.StatusCode(status).ExecuteAsync(ctx)
+                |> Async.AwaitTask
+                |> ignore
+            | _ -> failwith "Unexpected request."
+                
+        } |> Async.StartAsTask :> Task
+        )) |> ignore
+
+    let freePort =
+        let mutable port = 55555 // base listener port for the tests
+        while
+            IPGlobalProperties.GetIPGlobalProperties().GetActiveTcpListeners()
+            |> Array.map (fun x -> x.Port)
+            |> Array.contains port do
+                port <- port + 1
+        port
+
+    let baseAddress = $"http://localhost:{freePort}"
+
+    let workerTask = app.RunAsync(baseAddress)
+    printfn $"Started local http server with address {baseAddress}"
+
+    { new ITestHttpServer with
+        member this.Dispose() =
+            app.StopAsync() |> Async.AwaitTask |> ignore
+            printfn $"Stopped local http server with address {baseAddress}"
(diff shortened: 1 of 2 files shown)


tests/FSharp.Data.Tests/Http.fs
```diff
@@ -8,22 +8,85 @@ open System.Net
...
+            printfn $"Stopped local http server with address {baseAddress}"
+        member this.WorkerTask = workerTask
+        member this.BaseAddress = baseAddress }
 
 [<Test>]
 let ``Don't throw exceptions on http error`` () =
-    let response = Http.Request("http://httpstat.us/401", silentHttpErrors = true)
-    response.StatusCode |> should equal 401
+    let localServer = startHttpLocalServer()
+    using localServer <| fun _ ->
```
cartermp, line 74: Any reason not to do this?

```fsharp
use localServer = startHttplocalServer()
//...
```

tests/FSharp.Data.Tests/Http.fs
```diff
@@ -129,23 +192,31 @@ let ``Cookies is not added in cookieContainer but is still returned when addCook
...
 let ``Web request's timeout is used`` () =
-    let exc = Assert.Throws<WebException> (fun () ->
-        Http.Request("http://httpstat.us/200?sleep=1000", customizeHttpRequest = (fun req -> req.Timeout <- 1; req)) |> ignore)
-    Assert.AreEqual(typeof<TimeoutException>, exc.InnerException.GetType())
+    let localServer = startHttpLocalServer()
+    using localServer <| fun _ ->
+        let exc = Assert.Throws<WebException> (fun () ->
+            Http.Request(localServer.BaseAddress + "/200?sleep=1000", customizeHttpRequest = (fun req -> req.Timeout <- 1; req)) |> ignore)
+
+        Assert.That(exc.Status, Is.EqualTo(WebExceptionStatus.Timeout))
```
cartermp, line 200: Any reason why `Assert.AreEqual` can't still be used?

cartermp (commented):
Looks pretty good overall, just some questions

tests/FSharp.Data.Tests/Http.fs
mlaily: I started with an implementation where the `IDisposable returned by `startHttpLocalServer()` didn't contain anything useful beside `Dispose()` and I was afraid `Dispose()` would be called too early if the value wasn't otherwise used. (I'm not sure when it goes out of scope in this situation)

This is less true now that the `BaseAddress` is used to create the http request.

Do you want me to change it? (and if you know about when `Dispose()` is called with seemingly unused `use` bindings, I'd be interested to know :))

tests/FSharp.Data.Tests/Http.fs
mlaily: Not really. I find this syntax more easy to read, that's all (There is less ambiguity which value is the expected value and which one is the actual value).  

I can change it to `Assert.AreEqual` if you feel like consistency with the rest of the tests is more important.

cartermp: Yes, please change it so that it's consistent with the rest.

mlaily: Ok. (actually, the rest of the file uses `... |> should equal ...` so I'm going to do the same)

tests/FSharp.Data.Tests/Http.fs
cartermp: It will only get called when the whole function is out of scope, see here: https://sharplab.io/#v2:DYLgZgzgPg9gDgUwHYAIDKBPCAXBBbAWACh5l0tc8A6AFQAsAnBAQwBMBLJAc1uYgGsIxYtgyIUASRoIcACWzY4aBAwBuKlAF5iKXSk50V7bJIAi7CHBgRmAI2AIdeuzgbMAxibz5bGgEJ8CACCrKxMEBAgKK6cXE66LthunijeeL4MKADqMAz8KjR8/FGFAsJEDia2zAxgKAAUAJRa8SgA3ihICADuktJyCkoq6pndxnSteqk+Gth0FlTmltYITS1EU5t6cAyc2GCoAEQYCAjYh5N6aRkocws5eQVFWiil/FQASgCuSPVgPygAPooAC0AD4Go1GpddNdZvMIFQAhBgqFwhAXsdTucUABfVrESooOprbQbPRfFEoOh4PAvaq1SEE8m6HZ7A4oQ7dOjMXAjC4slBspD7I7MJCsW4IgVTYWizniyV3HAwGCHIA

mlaily: Oh ok! So nothing too fancy going on. Thanks for the demo! 🙏
I should use sharplab more often...

(I'll update the commit)

cartermp (approved):
Thanks!

mlaily:
Thank you too!

PR #1462, opened 2022-08-17
split up FSharp.Data take 2

dsyme:
Continues #1457 after https://github.com/fsprojects/FSharp.Data/issues/1459 caused revert https://github.com/fsprojects/FSharp.Data/pull/1460

Will require this workaround in the TypeProvider SDK https://github.com/fsprojects/FSharp.TypeProviders.SDK/pull/388


dsyme:
I've manually tested that updating to the latest TPSDK with the workaround makes the packages usable.

It's not easy to automate this testing - our doc generation does consume the DLLs but not the packages we build.  (Consuming built packages within the build itself is always surprisingly painful with .NET, as you have to use a local soure and make sure you're really getting the package version you built)

dsyme:
Changed the version number to `6.0.1-beta001`

PR #1470, opened 2023-01-06
Change the namespace of the public "InferenceMode" enum

mlaily:
This is a breaking change, but I feel like it's needed...

Using `FSharp.Data.Runtime.StructuralInference` was a mistake (sorry, my bad). It should not be necessary to reference that namespace just to initialize a type provider...

I moved the enum to the `FSharp.Data` namespace instead, the same as the providers so it's readily available:
- Nothing has to be explicitly `open`ed to use the enum with a type provider.
- This will hopefully make the breaking change less painful since it is likely to still compile after the update.

I have no idea how much this new-ish enum is used in the wild, but since it's optional and a bit obscure, it might not impact too many users...

Since the next version will be the new major version 6.x and it's still currently in beta, I think this is a good opportunity to do this kind of thing now.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,7 @@
+### 6.0.1-beta002 - Jan 6 2023
+
+* Change the namespace of the `InferenceMode` enum from `FSharp.Data.Runtime.StructuralInference` to `FSharp.Data`.
+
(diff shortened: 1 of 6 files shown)


cartermp (approved):
Definitely fine with the change here.

PR #1472, opened 2023-01-15
Fix reading CSV from non seekable network stream

Lenne231:
When using the `FSharp.Data.Csv.Core` package to read massive CSV data from an AWS S3 bucket we had some issues because sometimes many rows got lost.

The problem seems to be related to the `StreamReader.Peek()` method that is used. When `StreamReader.Peek()` returns `-1` the whole reading stops with the current implementation. The [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.peek?view=net-7.0) states that `-1` is not only returned in case of the end of the stream, but also if the stream is not seekable and the stream doesn't read all data that was requested.

> An integer representing the next character to be read, or -1 if there are no characters to be read or if the stream does not support seeking.

This can be fixed by using `StreamReader.Read()` instead, because this method is blocking. I have also tried to add a unit test to simulate the problem with a non seekable stream that always reads just 1 byte when calling `Stream.Read(buffer, offset, count)`.

### DIFF
--- a/src/Csv/CsvRuntime.fs
+++ b/src/Csv/CsvRuntime.fs
@@ -65,15 +65,14 @@ module internal CsvReader =
         /// Reads multiple lines from the input, skipping newline characters
         let rec readLines lineNumber =
             seq {
-                match reader.Peek() with
+                match reader.Read() with
                 | -1 -> ()
                 | Char '\r'
                 | Char '\n' ->
-                    reader.Read() |> ignore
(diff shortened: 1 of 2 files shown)


cartermp:
@Lenne231 looks like you need to run the code formatter.

Lenne231:
@cartermp thanks! Done!

cartermp (approved):
Thanks!

PR #1476, opened 2023-03-01
Fix generated code of the json provider with `PreferDictionaries` when values are arrays

mlaily:
Before this PR, the generation is wrong:

image

image

This is because the records inferred originally in the arrays are normally supposed to have the name of the property declaring them (`123` and `789` in the screenshots), and two similar records with a different name are not merged.

This problem was taken care of by the original `PreferDictionaries` implementation when values are records (by dropping the record's name), but not when the values are arrays of records!

After this PR, the names of records nested in an array are also dropped, which yields the expected result, with records properly merged in a single type:

image

(PR of the original implementation for reference: https://github.com/fsprojects/FSharp.Data/pull/1430)

---

_Note: This PR contains only 2 commits, but is currently on top of the `rearrange-project-files` branch, from https://github.com/fsprojects/FSharp.Data/pull/1475 — I will rebase it on `master` when https://github.com/fsprojects/FSharp.Data/pull/1475 is merged._

EDIT: rebased and ready!

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 * Move common runtime utilities out of `FSharp.Data.Http` and into a new `FSharp.Data.Runtime.Utilities` assembly.
 * Add `aria-label` to the list of html attributes used to infer names of types provided by the HtmlProvider.
 * Enable TLS 1.2 when requesting http(s) samples from the type providers.
+* Fix generated code of the json provider with `PreferDictionaries` when values are arrays.
 
 ### 6.0.1-beta001 - Aug 18 2022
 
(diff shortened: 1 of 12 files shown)


mlaily:
The build is failing because the worldbank api is apparently down -_- 
```
Error: D:\a\FSharp.Data\FSharp.Data\tests\FSharp.Data.Tests\WorldBankProvider.fs(16,5): error FS3021: Unexpected exception from provided type 'FSharp.Data.WorldBankData+ServiceTypes+Indicators' member 'GetMethods': The type provider 'ProviderImplementation.WorldBankProvider' reported an error: Failed to request 'http://api.worldbank.org/v2/indicator?per_page=1000&format=json&page=3'. Error: System.Net.WebException: Timeout exceeded while getting response ---> System.TimeoutException: The operation has timed out.
```

cartermp (approved):
Thank you! Also appreciate the extensive test cases.

PR #1483, opened 2023-04-27
Csv Column Inference ignores newline character

bonjune:
Changed the string reader in the `parseHeaders` function to erase newline characters before it runs inferences using the input schema.

This will resolve the issue: https://github.com/fsprojects/FSharp.Data/issues/952 .

A test to check this change is added. The test passes.


cartermp (approved):
Interesting approach! I like it, will merge if tests pass.

nightroman:
I wonder, what about `"\r\n"`?

```
            use reader = new StringReader(schema.Replace("\n", ""))
```

bonjune:
Oh, that's a valid point. I will make a patch.

PR #1486, opened 2023-05-02
Fix: inference on multiline csv schema

bonjune:
This resolves the point raised at https://github.com/fsprojects/FSharp.Data/pull/1483#issuecomment-1529633129 by @nightroman

* Defined a dedicated function to read schema string to break it down into strings, separated by comma, skipping newline characters (CR/LF).
* Used `ResizeArray` to build an array of strings, so `.Length` changed to `.Count` when used.


cartermp (approved):
Thanks, sorry for delay in review here!

PR #1227, opened 2018-11-06
Adding HTML active patterns to expose internals of HtmlNode.

NatElkins:
Support for https://github.com/fsharp/FSharp.Data/issues/1102

Issue #1102, opened 2017-11-13
Make HtmlAttribute and HtmlNode accessible


nojaf:
I'm trying to create script that parse some html to another format.

```fsharp
#r "./packages/FSharp.Data/lib/net45/FSharp.Data.dll"

open FSharp.Data
open System

let input = "<h1>Test <strong>one</strong> two</h1><ul><li>one</li><li>two</li><li>three</li>"

let html = HtmlNode.Parse input

let processHtmlNode (node:HtmlNode) =
    match node with
    | HtmlElement(name, attributes, elements) ->
        sprintf "%s [%s] []" name attrs
    | HtmlText content ->
        sprintf "str \"%s\"" content
    | HtmlComment content -> 
        sprintf "comment  \"%s\"" content
    | HtmlCData content ->
        "CData is deprecated in HTML5"

List.head html
|> processHtmlNode
```

Running this in the repl gives me an error that `html.fsx(13,7): error FS1093: The union cases or fields of the type 'HtmlNode' are not accessible from this code location`.

I believe this is due to the internal keyword on [line 17](https://github.com/fsharp/FSharp.Data/blob/master/src/Html/HtmlParser.fs#L17) and [line 31](https://github.com/fsharp/FSharp.Data/blob/master/src/Html/HtmlParser.fs#L31).

Would it be ok if I sent in a PR to remove the internal keyword there? Or do I have other options here?

ovatsus:
Its' internal by design, so if we change the internal representation it's not a breaking change (we had that problem before with JsonValue). Though I guess we could expose an active pattern

njlr:
I believe this can now be closed?

```fsharp
#r "nuget: FSharp.Data, 6.4.0"

open FSharp.Data

let input = "<h1>Test <strong>one</strong> two</h1><ul><li>one</li><li>two</li><li>three</li></ul>"

let html = HtmlNode.Parse input

let rec processHtmlAttribute (attr : HtmlAttribute) =
  let (HtmlAttribute (k, v)) = attr
  sprintf "\"%s\", \"%s\"" k v

let rec processHtmlNode (node : HtmlNode) =
  match node with
  | HtmlElement(name, attrs, elems) ->
    let attrsS = attrs |> Seq.map processHtmlAttribute |> String.concat "; "
    let elemsS = elems |> Seq.map processHtmlNode |> String.concat "; "
    sprintf "%s [%s] [%s]" name attrsS elemsS
  | HtmlText content ->
[...]

PR #1499, opened 2023-12-15
FSharp.Data.Html suggested change to enable people to use the html parser without hiding representation without having to hack the code itself.

smoothdeveloper:
I'm looking to use the html parsing DSL without the encumbering that active pattern brings in terms of using those to "avoid breaking changes" in the DU definitions, costing in terms of tooling and ways to do the matching using it.

For now, this is just so people can make their own build by flipping a tag in the fsproj, but we may consider having something similar to https://github.com/fsprojects/FSharp.Data.SqlClient/blob/c6e48a47eb1d64c7de7596bae3cb50a548592357/src/SqlClient/DataTable.fs#L10 as default, and provide the same package with "HIddenRepresentation" suffix to the nuget package name, for those that feel their code will be more resilient to this library changing things.

related: #1227, #1102, https://github.com/fsharp/fslang-suggestions/issues/1341


dsyme (approved):

dsyme:
@pr-fix

PR #1500, opened 2023-12-17
fix missing space

Fxplorer:
Update CsvRuntime.fs


EverybodyKurts:
Hmm, the pipeline for testing in windows failed. A quick glance of the stack trace states:

```bash
Hint: Could not find a version in your paket.dependencies file, consider adding 'version 6.2.1' at the top of your dependencies file (D:\a\FSharp.Data\FSharp.Data\paket.dependencies).
```

I this pr, I updated the required dotnet version from 6.0 to 7.0. When inside a devcontainer running on net6, I could not get ionide to load... it's only when I bumped up the dotnet version could I get ionide to load.

Unfortunately, I think rolling back to net6 would break the vscode devcontainer environment.

dsyme (approved):

PR #1503, opened 2024-03-11
Fix #1453: Fix CsvProvider encoding provider registration

### LINKED ISSUE #1453: CsvProvider encoding is not recognized
Hello,

I'm trying to use CsvProvider with specified encoding, so the definition looks like this:

```fsharp
    type TradeRows = CsvProvider<csvDefinitionFile, Separators = ";", SkipRows = 3, HasHeaders = true, Encoding = "1250", Culture = "cs-CZ", IgnoreErrors = true>

``` 

but I'm getting following error during build:

> No data is available for encoding 1250. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method. 

Funny thing is this project works completely fine in Visual Studio, but I cannot build it with `dotnet build` or Rider.

I understand I have to register encodings with something like this:

```fsharp
Encoding.RegisterProvider CodePagesEncodingProvider.Instance
```
but how to do it so it is recognized by CsvProvider?

dsyme:
Fixes #1453

Fix CsvProvider encoding provider registration

### DIFF
--- a/src/FSharp.Data.DesignTime/Csv/CsvProvider.fs
+++ b/src/FSharp.Data.DesignTime/Csv/CsvProvider.fs
@@ -109,12 +109,8 @@ type public CsvProvider(cfg: TypeProviderConfig) as this =
                 // Register CodePagesEncodingProvider before getting encoding
                 let provider = CodePagesEncodingProvider.Instance
                 Encoding.RegisterProvider provider
-                try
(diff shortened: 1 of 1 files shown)


dsyme:
This is the right fix but an additional component needs to be referenced I believe, see https://learn.microsoft.com/en-us/dotnet/api/system.text.codepagesencodingprovider?view=net-8.0

PR #1524, opened 2025-03-04
Fix singularization of words ending in `ases`

bisen2:
This PR would resolve https://github.com/fsprojects/FSharp.Data/issues/1508.

It would, however, be a breaking change. Any consumers currently using any `*ases` words as singular `*asis` would need to update their type names after updating this package.

I'm not sure if this is actually a change we would want to merge given the breaking nature, but figured I would open the PR and see what the maintainers thought.

### DIFF
--- a/src/FSharp.Data.Runtime.Utilities/Pluralizer.fs
+++ b/src/FSharp.Data.Runtime.Utilities/Pluralizer.fs
@@ -50,6 +50,7 @@ let private tables =
 
                "cis", "ces"
                "us", "uses"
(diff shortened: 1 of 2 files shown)


dsyme:
@bisen2 I'm OK with this change.  Could you add a major version bump as part of the change?  Thanks

bisen2:
Sure thing. Do you want me to mark it as a beta or just go straight for the 7.0.1, @dsyme?

dsyme:
@bisen2 Just straight to 7.0.1

